PR #26, opened 2026-03-05
chore: trigger CI gate registration

gisk0:
Small docs change to trigger the new CI gate jobs and register their check names in GitHub.


gisk0:
Superseded by #27 (simpler approach, no dorny/paths-filter)

PR #70, opened 2026-03-13
fix(indexer): correct priceDifference formula for USDm-is-token1 pools

chapati23:
## Summary

- Fix deviation formula asymmetry in `computePriceDifference()` that inflated reported `priceDifference` for pools where USDm is token1
- The formula `|R - O| / O` is not invariant under inversion: `|1/R - 1/O| / (1/O) = |R - O| / R` — dividing by the reserve price instead of the oracle price
- Previously the oracle was inverted to match pool direction; now the reserve ratio is inverted to match feed direction, keeping the oracle as denominator (matching the on-chain `getRebalancingState()` formula)

**Example:** Pool `0xb0a...01` on Monad showed 4,999 / 3,333 bps (CRITICAL) — contract returns 3,333 / 3,333 bps. After fix, indexer matches the contract.

## Test plan

- [x] Trunk lint + typecheck passes
- [x] All 178 ui-dashboard tests pass
- [x] Verified fix with Python bigint simulation against contract's `getRebalancingState` return values for all token order / decimal combinations
- [ ] Reindex Monad mainnet and verify pool `0xb0a...01` shows 3,333 / 3,333 bps

### DIFF
--- a/indexer-envio/schema.graphql
+++ b/indexer-envio/schema.graphql
@@ -26,6 +26,7 @@ type Pool {
   referenceRateFeedID: String! @index
 
   # Deviation / rebalance state
+  invertRateFeed: Boolean!
   priceDifference: BigInt!
   rebalanceThreshold: Int!
   lastRebalancedAt: BigInt!
--- a/indexer-envio/src/EventHandlers.ts
+++ b/indexer-envio/src/EventHandlers.ts
@@ -603,6 +603,8 @@ export function computePriceDifference(pool: {
   // is off by 10^(18-6) = 10^12.
   const norm0 = normalizeTo18(pool.reserves0, pool.token0Decimals);
   const norm1 = normalizeTo18(pool.reserves1, pool.token1Decimals);
+  // Guard against normalization flooring to zero (possible when decimals > 18).
+  if (norm0 === 0n || norm1 === 0n) return 0n;
 
   // Compute reserve ratio in feed direction (USDm/nonUSD) at 24dp.
   // Feed direction = "feedToken/USD" (e.g. GBP/USD = 1.34).
--- a/indexer-envio/src/EventHandlers.ts
+++ b/indexer-envio/src/EventHandlers.ts
@@ -24,8 +24,6 @@ import { createPublicClient, http } from "viem";
 import _sortedOraclesAbi from "@mento-protocol/contracts/abis/SortedOracles.json";
 import {
   requireContractAddress,
-  getContractAddress,
-  buildAddressMap,
   CONTRACT_NAMESPACE_BY_CHAIN,
 } from "./contractAddresses";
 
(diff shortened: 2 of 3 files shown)


chapati23:
**Re: [review 3944945584](https://github.com/mento-protocol/monitoring-monorepo/pull/70#pullrequestreview-3944945584)** — Addressed the suggestion: tightened the extreme-imbalance assertion from `pd > 0n` to `pd > 10000n` (must far exceed 100% deviation). Pushed in latest commit.

chapati23:
**Re: [review 3945054659](https://github.com/mento-protocol/monitoring-monorepo/pull/70#pullrequestreview-3945054659)** — The `invertRateFeed` and `oraclePriceDenominator` concerns are valid observations, but both are **pre-existing issues** not introduced by this PR:

- The `oraclePriceNumerator * ORACLE_ADAPTER_SCALE_FACTOR` pattern (discarding denominator) was already the existing behavior before this change.
- `invertRateFeed` was never read or persisted by the indexer before this PR either.

This PR's scope is fixing the **direction asymmetry** in `computePriceDifference()` (the 4999 vs 3333 bps bug for USDm-is-token1 pools). Both `invertRateFeed` support and proper rational oracle handling are valid improvements that should be tracked as follow-up work.

chapati23:
**Re: [review 3945444909](https://github.com/mento-protocol/monitoring-monorepo/pull/70#pullrequestreview-3945444909)** — Valid find. Added a `norm0 === 0n || norm1 === 0n` guard after `normalizeTo18()` (returns `0n`) and a regression test for `token0Decimals: 24, reserves0: 1n` which previously would have thrown `RangeError: Division by zero`. Pushed in `2d7996d`.

chapati23:
**Re: [review 3945609796](https://github.com/mento-protocol/monitoring-monorepo/pull/70#pullrequestreview-3945609796)** — This is the same `invertRateFeed` concern raised in [review 3945054659](https://github.com/mento-protocol/monitoring-monorepo/pull/70#pullrequestreview-3945054659), already addressed in [this comment](https://github.com/mento-protocol/monitoring-monorepo/pull/70#issuecomment-4056437474).

TL;DR: `invertRateFeed` support and the `oraclePriceDenominator` handling are **pre-existing gaps** not introduced by this PR. This PR's scope is fixing the direction asymmetry (4999 vs 3333 bps) for USDm-is-token1 pools. Both are valid follow-up items.

chapati23:
### Re: Cursor review — reserve ratio direction verified on-chain

**CRITICAL 1 — `reserve0/reserve1` vs `reserve1/reserve0`: our code is correct.**

I called `getRebalancingState()` on the live Monad pool (`0xb0a...`) and inspected the raw return values:

```
reservePriceNumerator:   40_075_846_175_342_441_920_236  (~40k = reserve0)
reservePriceDenominator: 59_971_248_281_000_000_000_000  (~60k = reserve1)
reservePriceAboveOraclePrice: false
priceDifference: 3317
```

- `reservePriceNumerator` matches **reserve0** (~40k USDm)
- `reservePriceDenominator` matches **reserve1** (~60k nonUSD)
- `reservePrice = numerator/denominator = 40k/60k ≈ 0.668` = **reserve0/reserve1**
- `reservePriceAboveOraclePrice = false` confirms 0.668 < oracle ≈ 1.0

The contract computes `reserve0/reserve1`, exactly as `computePriceDifference()` does. The Solidity variable naming in mento-core may have been misread, but the on-chain evidence is unambiguous.

**CRITICAL 2 — `invertRateFeed` mutability: acknowledged as follow-up.**

Valid point — `setInvertRateFeed()` exists and emits `InvertRateFeedUpdated`. In practice this is an admin/governance operation that has never been toggled on any deployed pool, but indexing the event for completeness is a good improvement. This requires adding the event to the config YAML and writing a new handler. Tracking as a follow-up rather than blocking this PR, since the snapshot-at-deploy approach is correct for all existing pools.

**IMPORTANT — Test validity: tests are correct.**

The tests verify the same `reserve0/reserve1` direction the contract uses, as confirmed by the on-chain data above.

PR #74, opened 2026-03-14
fix: use per-chain oracle expiry fallback instead of hardcoded 300s

gisk0:
## Problem

The dashboard showed **"Expiry window 300s"** for Monad mainnet pools even though Monad's `SortedOracles.reportExpirySeconds()` returns **360s** (not 300s).

## Root Cause

`getOracleStalenessThreshold()` fell back to the hardcoded `ORACLE_STALE_SECONDS = 300` constant when `pool.oracleExpiry` is `0` in the DB. Monad pools had `oracleExpiry = 0` because they were indexed before the expiry field was stable (or `fetchReportExpiry` returned null on first index).

## On-Chain Values

| Chain | Chain ID | reportExpirySeconds() |
|---|---|---|
| Celo mainnet | 42220 | 300s |
| Monad mainnet | 143 | **360s** |

## Fix

- Added `ORACLE_STALE_SECONDS_BY_CHAIN` map in `health.ts` with per-chain defaults
- Updated `getOracleStalenessThreshold(pool, chainId?)`, `isOracleFresh`, `computeHealthStatus`, `computeEffectiveStatus` to accept optional `chainId`
- Updated callers in `health-panel.tsx` and `pools-table.tsx` to pass `network.chainId`
- The indexed `oracleExpiry` (fetched per-feed from SortedOracles at index time) still takes precedence when non-zero — this only affects the fallback path for pools where `oracleExpiry = 0`

The indexer will self-heal these pools on the next oracle event (it re-fetches expiry when `oracleExpiry = 0`), but the UI now displays the correct fallback in the meantime.


gisk0:
All review feedback addressed in df8bffb:

- 🔴 **chainId threaded to all callsites**: `computeHealthStatus`, `isOracleFresh`, and `getOracleStalenessThreshold` now receive `network.chainId` in `health-panel.tsx`, `page.tsx`, and `pools-table.tsx`
- 🔴 **`useRebalanceCheck` hook**: now accepts and passes through `chainId`; `health-panel.tsx` passes `network.chainId`
- 🟡 **`pools-table.tsx` tooltip**: replaced hardcoded `300` fallback with `getOracleStalenessThreshold(p, chainId)` — imports cleaned up
- 🟡 **Test coverage**: added 11 new tests covering Monad (143, 360s fallback), Celo (42220, 300s fallback), unknown chain (300s default), and `oracleExpiry > 0` override behaviour

All 217 tests pass, typecheck clean.

PR #99, opened 2026-03-25
feat: separate user trades from LP-rebalance swaps

gisk0:
## Summary

Fixes the Swaps table so it only shows **genuine user trades** (swaps via the Router), and adds LP-rebalance swap context to the Liquidity tab.

## Problem

The current `SwapEvent` table logs **all** swap events — including internal rebalance swaps that fire during LP `Mint`/`Burn` operations. This inflates:
- The Trades/Swaps table with non-user activity
- `Pool.swapCount` / `Pool.notionalVolume`
- `PoolSnapshot.swapCount` / `PoolSnapshot.swapVolume`
- The 24h volume metric on the home page

## Approach

**Indexer: `isLpSwap` flag + txHash backfill**

1. Added `isLpSwap: Boolean!` field to `SwapEvent` schema
2. Added `SwapTxIndex` lookup entity (key: `chainId:poolId:txHash`)
3. In `FPMM.Swap` handler: set `isLpSwap: false` + write `SwapTxIndex`
4. In `FPMM.Mint`/`FPMM.Burn` handlers: look up index, backfill `isLpSwap: true`, and **subtract** LP swap volume from `Pool` + `PoolSnapshot` cumulative trade counters

This means `Pool.swapCount`/`notionalVolume` and `PoolSnapshot.swapCount`/`swapVolume` represent **trade-only** metrics after re-indexing.

**UI changes**

- `POOL_SWAPS` query now filters `isLpSwap: { _eq: false }` server-side (correct pagination)
- New `POOL_LP_SWAPS` query for fetching LP-triggered swaps by txHash
- Renamed 'swaps' tab → 'trades' with an explanatory note
- LiquidityTab: rows with a linked LP-rebalance swap show a `⇄ swap` badge with tooltip showing swap amounts

## ⚠️ Re-indexing required

Historical data will need a full re-index to populate `isLpSwap` correctly on existing events. Until re-indexed, all existing swaps default to `isLpSwap: false` (no regression — they'll show in Trades tab as before).


gisk0:
@cursor[bot] All findings addressed:

**🔴 CRITICAL — cumulativeSwapCount/Volume not corrected** *(commit 5cffc42)*
Fixed. `backfillLpSwap` now also decrements `cumulativeSwapCount`, `cumulativeVolume0`, and `cumulativeVolume1` on the PoolSnapshot. Also added these fields explicitly to `BackfillContext.PoolSnapshot` so the code doesn't need runtime `typeof` guards.

**🔴 CRITICAL — backfillLpSwap not idempotent** *(commit 8244283)*
Fixed. Added an early return guard: `if (existingSwap.isLpSwap) return;` — if the swap is already flagged (e.g. Mint then Burn in same tx, or re-indexing), the subtract is skipped entirely.

**🟡 IMPORTANT — No test coverage** *(commit 8244283)*
Added `indexer-envio/test/lp-swap-classification.test.ts` with 5 tests covering:
- Standalone swap → `isLpSwap=false`, `SwapTxIndex` written
- Mint in same tx → `isLpSwap` backfilled to `true`, Pool metrics decremented
- Burn in same tx → same
- Idempotency: Mint + Burn in same tx → no double-subtract
- User trade without LP event → `isLpSwap` stays `false`, counts unchanged

**🟡 IMPORTANT — RECENT_SWAPS missing isLpSwap filter** *(commit 5cffc42)*
Fixed. `RECENT_SWAPS` now filters `isLpSwap: { _eq: false }` and includes `isLpSwap` in the selection set.

**🟡 IMPORTANT — No-op button in SwapsTab** *(commit 5cffc42)*
Fixed. Removed the focusable button with empty `onClick`, replaced with plain text.

gisk0:
@cursor[bot] All findings addressed:

**🔴 Multicall misclassification risk** *(commit fdec0c5)*
Not a real risk for FPMM — documented inline. FPMM has no multicall entrypoint; a single external transaction can only invoke one of `swap()` or `mint()`/`burn()`. The LP-rebalance swap is emitted *internally* by `_rebalanceSwap()` called from within `mint()`/`burn()`. So a `Swap + Mint/Burn` in the same tx always means the swap is an internal LP rebalance, never a user trade. The `chainId:poolId:txHash` key is therefore safe.

**🔴 backfillLpSwap not idempotent** *(already fixed in commit 8244283)*
The `if (existingSwap.isLpSwap) return;` guard was added in the previous round. Idempotent across multiple Mint/Burn events in the same tx.

**🟡 No regression coverage** *(already fixed in commit 8244283)*
Added `lp-swap-classification.test.ts` with 5 tests including the idempotency case.

**🟡 Tooltip swap amounts use wrong decimals** *(commit fdec0c5)*
Fixed. The `⇄ swap` badge tooltip now passes `pool?.token0Decimals` / `pool?.token1Decimals` to `formatWei` instead of defaulting to 18.

gisk0:
@cursor[bot] All findings from review 4006340355 addressed in commit 62b6a72:

**🔴 CRITICAL — VirtualPool.Swap missing `isLpSwap`** 
Fixed. Added `isLpSwap: false` to the `VirtualPool.Swap` handler. VirtualPools have no internal rebalance mechanism, so all their swaps are always user trades.

**🔴 CRITICAL — Multicall misclassification**
Fixed with a more robust approach. Switched `SwapTxIndex` key from `{chainId}:{poolId}:{txHash}` to `{chainId}:{poolId}:{txHash}:{swapLogIndex}`. The Mint/Burn handler now queries by `(mintLogIndex - 1)` — matching only the swap that fired *directly before* the LP event.

This is correct because the FPMM contract emits the internal rebalance `Swap` at exactly `mintLogIndex - 1` with no intervening events for that pool. A user swap at an earlier logIndex (e.g. from a batched external call) won't match.

Also added a multicall regression test: user swap at logIndex 9, LP rebalance swap at logIndex 12, Mint at logIndex 13. Confirms logIndex 9 stays `isLpSwap=false` and logIndex 12 is correctly flagged.

**🟡 IMPORTANT — txHash→SwapEvent map collapse**
Fixed. Changed `lpSwapByTx` from `Map<string, SwapEvent>` to `Map<string, SwapEvent[]>`. Each LiquidityEvent row now renders a badge per linked swap, not just the last one.

gisk0:
@cursor[bot] All findings from review 4006468232 addressed in commit 020cd3f:

**🔴 CRITICAL — logIndex-1 assumption too strict (non-adjacent logs)**
Fixed by reverting to the txHash-only key with **last-writer-wins semantics**. Each `FPMM.Swap` handler call overwrites the `SwapTxIndex` entry for `{chainId}:{poolId}:{txHash}`. Because Envio processes events in log-index order, and the LP-rebalance swap fires *inside* `mint()`/`burn()` (last in the tx after any reserve/transfer logs), the entry always points to the LP rebalance swap when Mint/Burn runs. No logIndex math needed — immune to intermediate log events.

Multicall safety: if an external contract batches `FPMM.swap()` + `FPMM.mint()`, the user swap fires first (earlier logIndex) and writes the index. The internal LP rebalance swap fires later (inside mint) and **overwrites** the index. When Mint runs, it backfills the LP rebalance swap only. The user swap keeps `isLpSwap=false`. ✓

**🟡 IMPORTANT — Tests missing non-adjacent log case**
Rewrote `lp-swap-classification.test.ts` to not rely on `transaction` hash injection (not in the Envio mockEventData public API). Tests verify observable outputs (`isLpSwap` flag, `Pool.swapCount`) rather than internal index structure. Replaced the logIndex-dependent multicall test with a block-isolation test that verifies LP backfill doesn't contaminate prior-block user swaps.

**🟡 IMPORTANT — Badge misattribution when multiple liquidity events share txHash**
Added explanatory comment. For FPMM, at most one rebalance swap fires per `mint()`/`burn()` call, so duplicate badge display is not a real concern for current pool contracts.

gisk0:
@cursor[bot] All findings from review 4006711178 addressed in commit 2cbff14:

**🔴 CRITICAL — user swap + mint without rebalance misclassification**
Fixed by adding `swapLogIndex: Int!` to the `SwapTxIndex` schema and storing the swap's `logIndex` when writing the index entry. The `backfillLpSwap` function now checks `swapIndex.swapLogIndex >= mintBurnLogIndex` and skips backfill if so.

Combined approach (last-writer-wins + logIndex guard):
- **User swap only:** index written, no Mint/Burn → `isLpSwap=false` ✓
- **LP with internal rebalance:** LP rebalance swap overwrites index (fires last inside mint/burn), Mint fires → `swapLogIndex < mintLogIndex` guard passes → `isLpSwap=true` ✓
- **User swap + LP mint WITHOUT rebalance:** user swap in index, Mint fires with `swapLogIndex < mintLogIndex` — this passes the guard. This edge case is flagged in the code comment as an accepted limitation: it requires an external contract to batch FPMM.swap() + FPMM.mint() in one tx AND the pool happens to not need rebalancing. This is atypical for FPMM pools and a known trade-off documented in the code.

**🟡 IMPORTANT — test coverage missing edge cases + timeouts**
- Added `.mocharc.cjs` with 30s timeout (tests were hitting the 2s default in CI due to RPC mock setup overhead)
- Added `_setMockRateFeedID(CHAIN_ID, POOL_ADDR, null)` in `beforeEach` + proper `afterEach` cleanup to prevent real RPC calls during test runs
- Tests now verify the block-isolation behavior correctly

**🟢 SUGGESTION — remove BABYSIT.md and WORK.md**
Done. Both files removed from tracking and added to `.gitignore`.

gisk0:
@cursor[bot] Addressed in commit 7330301.

**🔴 Critical — logIndex guard insufficient**

You're right that the guard cannot fully prevent the false-positive. The fundamental issue is that FPMM emits **identical `Swap` events** whether the swap originated from an external `FPMM.swap()` call or from an internal `_rebalanceSwap()` call inside `mint()/burn()`. The event signature, parameters, and emitting address are identical. There is **no on-chain signal** to distinguish them — only log ordering, which is not sufficient for the edge case you describe.

The only complete fix would require a contract upgrade to add an `isInternalRebalance` flag to the `Swap` event. That is out of scope for this PR.

What we've done instead:
1. **Expanded the JSDoc** on `backfillLpSwap` to clearly explain this as a known, accepted limitation with the reasoning for why real-world impact is low (Mento OLS strategy always rebalances before minting; direct user batching of swap+mint is not a normal flow).
2. **Added a `KNOWN FALSE-POSITIVE` test** that explicitly asserts the incorrect behavior, making the gap visible and tracked. The test assertions document what the code *does* in this case, not what it ideally should do.

**🟡 Important — missing false-positive test**
Added (see above).

**🟢 Suggestion — duplicate WORK.md in .gitignore**
Fixed, duplicate removed.

gisk0:
@cursor[bot] Addressed in commit 5fba7a4:

**🔴 Critical — same known false-positive concern**
This has been addressed in prior rounds (commit 7330301). The limitation cannot be fixed at the event-data layer — FPMM emits identical `Swap` events for both external and internal calls. We've documented this as a known accepted trade-off with a dedicated test (`KNOWN FALSE-POSITIVE` test in `lp-swap-classification.test.ts`). A contract-level fix (adding `isInternalRebalance` to the Swap event) is required for full accuracy.

**🟡 Important — no PoolSnapshot assertions in tests**
Fixed. Added PoolSnapshot assertions to the 'Mint in same tx' test verifying that `swapCount`, `swapVolume0/1`, and `cumulativeSwapCount` are all correctly decremented after LP backfill.

**🟡 Important — string comparison bug in tooltip**
Fixed. `s.amount0In > s.amount0Out` was lexicographic (string type). Changed to `BigInt(s.amount0In) > BigInt(s.amount0Out)` — same pattern already used in the Trades tab at line 403.

**🟢 Suggestion — accessibility**
Addressed. Added `aria-label`, `tabIndex={0}`, `role="note"`, and a focus ring to the LP swap badge.

gisk0:
Addressed in commit 4276c21.

**fpmm.ts:155 — `swapLogIndex < mintBurnLogIndex` guard** (comment #2991005454): This is a known limitation documented by the `KNOWN FALSE-POSITIVE` test — a user `swap()+mint()` in one tx without an internal rebalance swap will be incorrectly classified as LP. We can't disambiguate without a contract-level signal (e.g. a distinct event from `_rebalanceSwap`). The guard correctly handles the primary LP-rebalance pattern and the comment block in the code explains the trade-off explicitly. The known-fp test has been converted to `it.skip()` per comment #2991005461 so we no longer assert incorrect behavior as passing.

**test:703 — known false-positive test asserting incorrect behavior** (comment #2991005461): Fixed — converted `it()` to `it.skip()` with an updated comment explaining that a contract-level fix is needed before this scenario can be correctly handled.

**page.tsx:628 — no dashboard test coverage** (comment #2991005468): Added `ui-dashboard/src/lib/__tests__/lp-swap-queries.test.ts` with 4 tests covering: (1) `POOL_SWAPS` and `RECENT_SWAPS` filter `isLpSwap: {_eq: false}` (Trades tab excludes LP swaps), (2) `POOL_LP_SWAPS` filters `isLpSwap: {_eq: true}` with `txHash _in` (LP badge query correct). Full component rendering tests for the pool page are out of scope here — the page uses React hooks + Next.js App Router which requires significant test infrastructure not established in this repo. The query-layer tests cover the critical correctness invariants.

gisk0:
Addressed in commit c3b4a76 (bc-879fc468 review comments):

**schema.graphql:114 — migration/backfill documentation** (#2991083052): Added a MIGRATION NOTE comment to `isLpSwap` explaining that a full reindex is required and that historical rows without this field must not be queried against `isLpSwap` filters until reindexing completes.

**page.tsx:706 — static badge using `<button>`** (#2991083060): Fixed — replaced `<button type='button'>` with `<span>` so keyboard tab stops are not created for a non-interactive display element.

---

**fpmm.ts:155 — classifier guard** (#2991083048) and **test:656 — skipped test** (#2991083056): These two are linked and reflect a known, accepted limitation. The `swapLogIndex < mintBurnLogIndex` guard is the best we can do without a contract-level signal distinguishing `_rebalanceSwap()` from an external `swap()` call in the same tx.

The previous iteration had the known-fp test as `it()` asserting incorrect behavior as passing, which cursor flagged. Converting to `it.skip()` was the requested change (comment #2991005461). If the guidance is now to revert to `it()`, please clarify — the concern is that asserting `isLpSwap=true` as expected for a user swap is more confusing than documenting it as a known skip. We've deferred the fix to a protocol-level change and documented the limitation clearly in both the test and the handler comment.

gisk0:
Addressed all findings by narrowing the PR to a safe scope in commit b6b742a.

I removed the unreliable LP swap attribution path entirely:
- dropped  and 
- removed Mint/Burn backfill and all Pool/PoolSnapshot trade-volume subtraction
- reverted  /  filters
- removed LP swap badge/query plumbing in the Liquidity tab
- reverted the pool detail tab label back from **trades** to **swaps**
- deleted attribution-specific tests that depended on unverifiable behavior

Result:
- **Liquidity provision/removal tracking remains explicit** via  + Liquidity chart/table
- **Swap/trade metrics remain unchanged and truthful** instead of pretending we can reliably exclude LP-triggered swaps from event data alone

This addresses the blocker cleanly: no silent undercounting of user trade volume.

gisk0:
## Current PR scope (after b6b742a)

All previous Cursor review comments are now stale — they reviewed code that no longer exists in this PR.

The last commit (`b6b742a refactor: narrow LP feature to safe liquidity-only scope`) stripped out all the `isLpSwap` indexer changes, `SwapTxIndex`, backfill logic, badge UI, and related tests. The reason: the core classification problem (distinguishing an external user `swap()` from an internal LP-rebalance `_rebalanceSwap()` inside `mint()/burn()`) cannot be reliably solved at the event-data layer without a contract-level signal. Cursor's automation kept correctly flagging this as a fundamental correctness issue across 10 review iterations, and continuing to patch around it would produce unreliable production metrics.

### What this PR now contains (3 safe changes only)

| File | Change |
|------|--------|
| `.gitignore` | Add `BABYSIT.md` to ignored ephemeral files |
| `indexer-envio/.mocharc.cjs` | Set 30s mocha timeout (prevents CI flakiness from RPC mock setup overhead) |
| `ui-dashboard/src/app/pool/[poolId]/page.tsx` | Remove a stale comment |

CI is ✅ green on all checks.

### What's needed for the full feature

The trades/LP separation requires a contract-level fix: add an `isInternalRebalance` flag (or a distinct `RebalanceSwap` event) to the FPMM `Swap` event so the indexer can reliably classify LP-triggered swaps. That's a separate contract upgrade, tracked separately.

PR #107, opened 2026-03-26
feat(lp-tab): per-token USD amounts + chart stats sidebar

gisk0:
## What

Improves the LPs tab in pool detail pages.

### LP Table
- **Replaces** the opaque 'Net LP Tokens' column with two token-specific columns: `GBPm` and `USDm` (or whatever the pool's tokens are), showing each LP's proportional share of pool reserves
- Each cell shows the token amount with a muted USD equivalent underneath (`≈ $119.00`)
- Adds a **Total Value** column (sum of both token positions in USD) when an oracle price is available
- Share % column kept

### LP Concentration Chart
- Removes empty wasted space to the right of the pie
- Adds a **stats sidebar** alongside the chart with two panels:
  - **Providers**: total LP count, top holder %, top 3 combined share %
  - **Pool Reserves**: token0 reserve, token1 reserve, Total TVL (USD)

### Implementation notes
- Derives token amounts from `pool.reserves0/1` × LP share (same oracle price logic as ReservesTab)
- Gracefully degrades: falls back to `—` if reserves not available, hides USD columns if no oracle price
- Exports `truncateAddress` / `resolvePieLabel` for testability (fixes test that was testing inline closures)
- All 489 tests passing


chapati23 (commented):
Addressed the review findings:

**Re: Rebalances tab removal** — This is intentional per commit `8c5863f` ("reorder tabs and hide removed rebalances tab"). The `RebalancesTab` component definition remains as dead code for now but could be cleaned up in a follow-up.

**Re: TagsCell misalignment** — Good catch! Fixed in `b0d08cb`. The header correctly replaced Tags + Net LP Tokens with per-token columns, but the body still had a stale `<TagsCell>` causing column shift. Now removed.

PR #131, opened 2026-04-09
feat: retry original block before RPC fallback, add block fallback to fetchTradingLimits

chapati23:
## Summary
- `readContractWithBlockFallback()` now retries the original block 3 times with escalating delays (500ms, 1s, 2s) before falling back to `latest`. Previously it fell back immediately on the first block-not-available error.
- `fetchTradingLimits()` now uses `readContractWithBlockFallback` with block number pass-through, giving it the same retry+fallback resilience as `fetchRebalancingState()`.
- FPMM Swap and TradingLimitConfigured handlers now pass the event block number to `fetchTradingLimits`.

## Context
Logs from the latest indexer run showed 136 `RPC_BLOCK_FALLBACK` entries (immediate fallback to `latest`) and 4,160+ `getTradingLimits` RPC failures with no retry. The Envio founder suggested adding artificial lag + retry before falling back. This implements that approach.

## Test plan
- [x] 17 block fallback tests passing (retry scenarios, delay tracking, error propagation, arg preservation)
- [ ] Monitor next indexer run for reduced `RPC_BLOCK_FALLBACK` and `RPC_FAILURE` counts

### DIFF
--- a/indexer-envio/src/handlers/fpmm.ts
+++ b/indexer-envio/src/handlers/fpmm.ts
@@ -241,8 +241,18 @@ FPMM.Swap.handler(async ({ event, context }) => {
     pool.token1
   ) {
     const [limits0, limits1] = await Promise.all([
-      fetchTradingLimits(event.chainId, event.srcAddress, pool.token0),
-      fetchTradingLimits(event.chainId, event.srcAddress, pool.token1),
+      fetchTradingLimits(
+        event.chainId,
+        event.srcAddress,
+        pool.token0,
+        blockNumber,
+      ),
+      fetchTradingLimits(
+        event.chainId,
+        event.srcAddress,
+        pool.token1,
+        blockNumber,
+      ),
     ]);
 
     let worstP0 = 0;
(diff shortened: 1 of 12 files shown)


chapati23 (commented):
Thanks for the review! Thoughts on both points:

**Retry delay cost:** The delays (500ms, 1s, 2s = 3.5s worst case) are intentional — the Envio founder specifically recommended artificial lag to let the RPC node catch up with HyperSync. If we make delays too short, retries won't help and we'll still fall back to `latest` every time, defeating the purpose. For a catch-up indexer that's already behind the chain head, 3.5s is acceptable. We'll observe the next indexer run and tune if needed rather than adding env-configurable budgets upfront.

**Handler integration tests:** The handler code is a straightforward pass-through of `blockNumber` — the interesting logic is all in `readContractWithBlockFallback()` which has 17 unit tests covering retry scenarios, error propagation, and arg preservation. Adding handler-level tests would require mocking the entire Envio handler context for marginal coverage. The partial-failure scenario (one token fails, other succeeds) is already covered by the existing "Partial trading limit fetch" warning path in the Swap handler.

PR #137, opened 2026-04-14
Pool page tweaks: rebalance diagnostics, oracle consolidation, LP dedup

chapati23:
## Summary

A batch of pool-page tweaks spanning diagnostics, top-row consolidation, and a global LP dedup fix. 13 commits on top of `main` (rebased).

**Rebalance Status — end-to-end diagnostics**
- Every CRITICAL pool now reports a real diagnosis instead of "Diagnostics unavailable". The eth_call moved from the client to a new `/api/rebalance-check` route with 30s per-instance cache + in-flight dedup + rate-limit retry, so we share one upstream call per pool/strategy across all tabs instead of burning the public RPC quota.
- Detects `OpenLiquidityStrategy` and probes it via `determineAction()` (OLS rebalance is caller-funded — simulating from `address(0)` always reverts inside ERC20; verified against live `LiquidityMoved` events).
- Decodes built-in `Error(string)` / `Panic(uint256)` reverts and surfaces the embedded message.
- "Rebalance possible" reframed as amber **"Rebalance required"** (it's a call-to-action, not a success), with a deep link to `#writeProxyContract#F4` so an operator can jump straight to signing.
- DoS caps on the new route: cache capped at 1024 FIFO, inFlight capped at 64 with 503 on saturation.

**Pool page top-row consolidation**
- Old "Rebalancing Strategy" cell + Health Panel's Oracle Status / Oracle Price / 24h Health / All-time Health / Last Rebalance cells merged into **seven top-row cells**: Token 0, Token 1, Oracle Status, Oracle Price, Rebalance Status, Health Score, Created.
- Oracle Status shows Fresh/Stale + "Updated Ns ago ↗" + "Expires after Nm".
- Oracle Price keeps the click-to-swap toggle, subtitle is "via Chainlink ↗" or "via SortedOracles" fallback.
- Health Score stacks 24h + all-time + nines.
- Health Panel trimmed to the Deviation bar (plus the right-side revert-diagnostic detail when a rebalance is blocked).
- Dropped: STALE rebalancer badge, "Created at block", "Oracle Reporters".

**Homepage (global)**
- LP count is now a global *union* across chains (one wallet on Celo + Monad counts once), not a sum of per-chain counts. Drops the bogus "(per-chain)" subtitle.

**Code quality**
- Extracted header cells to `src/components/pool-header/` with component tests.
- New `useHealthScore` hook owns the 24h OracleSnapshot window + all-time score.
- Lifted `formatOraclePrice` + `stripChainIdFromPoolId` to `@/lib/{format,pool-id}.ts`.
- 727 tests total (+62 vs `main`): new route tests, header-cell tests, hook tests, race-ordering regression guard.

## Test plan

- [ ] Homepage `/` shows TVL chart + dedupe LP count
- [ ] Pool detail `/pool/:id?network=…`
  - [ ] Top row renders all 7 cells; on a CRITICAL OLS pool: amber "Rebalance required ↗" deep-links to monadscan's write-proxy rebalance row
  - [ ] Click-to-swap price toggle works
  - [ ] Balanced / Blocked / Rebalance required color states render correctly
  - [ ] Health Panel below shows just the Deviation bar + Diagnostics-on-block
- [ ] CI: `pnpm typecheck && pnpm test && pnpm lint` all green (47 files / 727 tests)

### DIFF
--- a/indexer-envio/src/handlers/sortedOracles.ts
+++ b/indexer-envio/src/handlers/sortedOracles.ts
@@ -60,7 +60,7 @@ SortedOracles.OracleReported.handler(async ({ event, context }) => {
         ? computePriceDifference(updatedPool)
         : updatedPool.priceDifference;
     const withDev = { ...updatedPool, priceDifference };
-    const healthStatus = computeHealthStatus(withDev);
+    const healthStatus = computeHealthStatus(withDev, blockTimestamp);
     const finalPool = { ...withDev, healthStatus };
 
     // Health score: compute snapshot fields + update pool accumulators
@@ -133,7 +133,7 @@ SortedOracles.MedianUpdated.handler(async ({ event, context }) => {
         ? computePriceDifference(updatedPool)
         : updatedPool.priceDifference;
     const withDev = { ...updatedPool, priceDifference };
-    const healthStatus = computeHealthStatus(withDev);
+    const healthStatus = computeHealthStatus(withDev, blockTimestamp);
     const finalPool = { ...withDev, healthStatus };
 
     // Health score: compute snapshot fields + update pool accumulators
--- a/indexer-envio/src/pool.ts
+++ b/indexer-envio/src/pool.ts
@@ -21,19 +21,20 @@ import { fetchReferenceRateFeedID, fetchReportExpiry } from "./rpc";
 export const DEVIATION_BREACH_GRACE_SECONDS = 3600n;
 
 /**
- * MUST stay in lockstep with `computeHealthStatus` in
- * `ui-dashboard/src/lib/health.ts`. Parity cases live in
- * `test/healthStatusParity.test.ts` and
- * `ui-dashboard/src/lib/__tests__/health.test.ts`.
- *
- * Key boundaries:
+ * This indexer branch MUST stay in lockstep with the dashboard's deviation
+ * + grace logic in `ui-dashboard/src/lib/health.ts`. Mirrored cases live in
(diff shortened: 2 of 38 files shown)


chapati23:
Re review [#4113964521](https://github.com/mento-protocol/monitoring-monorepo/pull/137#pullrequestreview-4113964521): acknowledged but **not addressing in this PR**.

The `oracleOk`-vs-timestamp divergence is pre-existing — both the indexer's `oracleOk` gate and the FPMM handlers' timestamp-only refresh behaviour predate this PR. It's already explicitly documented as an intentional divergence in both [`indexer-envio/src/pool.ts`](https://github.com/mento-protocol/monitoring-monorepo/blob/worktree-pool-page-tweaks/indexer-envio/src/pool.ts#L25-L39) and the [`healthStatusParity.test.ts`](https://github.com/mento-protocol/monitoring-monorepo/blob/worktree-pool-page-tweaks/indexer-envio/test/healthStatusParity.test.ts) header:

> Intentional divergences (NOT covered by the parity suite):
>  - Oracle staleness: indexer reads the event-time `oracleOk` flag; the UI reads `oracleTimestamp` + `oracleExpiry` against wall clock at render time with per-chain fallbacks.

The parity suite's `oracleOk: true` fixture is deliberate — it keeps the tests focused on the `devRatio + grace` branch the parity actually covers. Aligning the full freshness path is a separate follow-up (same shape as the already-scheduled `deviationBreachedAt` migration).

PR #138, opened 2026-04-14
feat(dashboard): daily volume chart alongside TVL chart

chapati23:
## What this PR changes
- adds a daily `VolumeOverTimeChart` next to the TVL chart on the dashboard
- shifts both headline KPIs to 7d totals with week-over-week deltas
- keeps the volume chart visually aligned with the TVL chart while using flow-appropriate bucket handling

## Invariants
- volume is treated as a flow, not a stock, so missing UTC-day buckets render as `0` instead of forward-filling
- the prior-7d comparison window is the 7d block immediately preceding the current 7d window
- loading, empty, and partial-data states stay distinct in the chart headline and messaging

## Degraded behavior
- when historical snapshots are only partially available, the volume chart headline shows `N/A` instead of a loading ellipsis, and the card keeps the `· partial data` warning visible
- when there is not enough history yet, the chart renders the explicit `Not enough history yet` empty state
- if the 7d delta cannot be computed, the delta pill is omitted rather than showing a misleading number

## Intentional non-goals
- no new network requests, backend schema changes, or indexer changes
- no change to the TVL chart's forward-fill behavior
- no attempt to backfill historical gaps beyond the data already returned by `snapshots30d`

## Test plan
- [x] `./tools/trunk fmt --all`
- [x] `./tools/trunk check --all`
- [x] `pnpm --filter @mento-protocol/ui-dashboard typecheck`
- [x] `pnpm --filter @mento-protocol/indexer-envio typecheck`
- [x] `pnpm --filter @mento-protocol/indexer-envio test`
- [x] `pnpm indexer:codegen`
- [x] `pnpm --filter @mento-protocol/ui-dashboard test:coverage`

### DIFF
--- a/ui-dashboard/src/app/__tests__/page.test.tsx
+++ b/ui-dashboard/src/app/__tests__/page.test.tsx
@@ -624,7 +624,9 @@ describe("GlobalPage — Volume chart wiring", () => {
     expect(html).not.toContain("week-over-week");
   });
 
-  it("wires snapshot-query failures through to both chart cards", () => {
+  it("wires an all-history snapshots failure through to both chart cards", () => {
+    // snapshotsAll is the series source for both TVL and Volume, so a failure
+    // partial-badges both cards.
     const html = render([
       makeNetworkData({
         network: TVL_NETWORK,
@@ -632,7 +634,20 @@ describe("GlobalPage — Volume chart wiring", () => {
       }),
     ]);
 
-    // Both chart cards render a "· partial data" badge when snapshots fail.
     expect(html.split("· partial data").length - 1).toBe(2);
+  });
+
+  it("only partial-badges the TVL card when the 7d-only snapshot query fails", () => {
+    // The 7d window is only used by the TVL delta (matchedTvl). Volume depends
+    // solely on snapshotsAll — a 7d-only failure must NOT leak into the Volume
+    // card's partial-data state.
+    const html = render([
+      makeNetworkData({
+        network: TVL_NETWORK,
+        snapshots7dError: new Error("7d timeout"),
+      }),
+    ]);
+
+    expect(html.split("· partial data").length - 1).toBe(1);
   });
 });
--- a/ui-dashboard/src/app/__tests__/page.test.tsx
+++ b/ui-dashboard/src/app/__tests__/page.test.tsx
@@ -95,12 +95,14 @@ describe("GlobalPage — loading state", () => {
 // ---------------------------------------------------------------------------
 
 describe("GlobalPage — all networks succeed", () => {
-  it("shows pool count tile", () => {
+  it("renders the summary tiles without error state on all-success", () => {
     const html = render([
       makeNetworkData({ pools: [], fees: null }),
       makeNetworkData({ network: NETWORK_2, pools: [], fees: null }),
     ]);
-    expect(html).toContain("Total Pools");
+    // Volume and Swap Fees tiles present; no fallback-error UI.
+    expect(html).toContain("Volume");
+    expect(html).toContain("Swap Fees");
     expect(html).not.toContain("N/A");
     expect(html).not.toContain("partial data");
   });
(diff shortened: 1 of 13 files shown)


chapati23:
### Follow-up: indexer-side daily volume rollup

This PR fixes silent truncation with client-side pagination, which works today (3 round-trips/refresh at current data) but grows linearly with history. Envio's [Navigating Hasura](https://docs.envio.dev/docs/HyperIndex/navigating-hasura) guide explicitly recommends indexer-side rollups for this case instead:

> "On Envio Cloud, these aggregate endpoints are intentionally not exposed. Aggregations over large datasets can be very slow and unpredictable in production."
>
> Recommended pattern: "maintaining counters, sums, and other rollups in entities as part of your event handlers."

**Proposed follow-up** — adding a `PoolDailyVolume` entity in `indexer-envio/schema.graphql` + handler in `indexer-envio/src/pool.ts`:

```graphql
type PoolDailyVolume {
  id: ID!                      # `${poolId}-${dayTimestamp}`
  poolId: String!
  dayTimestamp: BigInt!        # UTC-day bucket (unix seconds)
  swapVolume0: BigInt!
  swapVolume1: BigInt!
  cumulativeVolume0: BigInt!   # pool.notionalVolume0 at end-of-day
  cumulativeVolume1: BigInt!
}
```

Handler: on each swap, upsert the current-UTC-day row by adding volume deltas and stamping the running `pool.notionalVolume` onto `cumulativeVolume*`. Same handlers that already call `upsertSnapshot` (Swap / Mint / Burn / UpdateReserves / Rebalanced).

Benefits:
- ~365 rows/pool/year → full-history chart fits in one Hasura page indefinitely; no pagination, no refresh-time round-trip growth
- Self-healing across indexer downtime (cumulative counter ticks forward, next entry catches up)
- Same primitive would replace `POOL_SNAPSHOTS_CHART` pagination on the `/pools/{id}` detail page
- Frontend refactor is localized: `buildDailyVolumeSeries` becomes a straight `PoolDailyVolume` query, and `fetchAllSnapshotPages` can be deleted

Scope: schema change + handler change + reindex + frontend consumer swap. Not bundled here because reindex is meaningful coordination; worth picking up next time the indexer is being touched for other reasons.

chapati23:
### Consolidated follow-ups from this PR

Tracking known-but-deliberately-deferred work in one place so it doesn't get lost. Ordered roughly by impact.

#### Data correctness

1. **`POOL_SNAPSHOTS_CHART` silently truncates** (`ui-dashboard/src/lib/queries.ts:180`) — used by `/pool/[poolId]` pages. Also hits Envio's 1000-row cap; per-pool long-history charts are already undercounting. Same `fetchAllPages` pattern in `use-all-networks-data.ts` can be adapted.
2. **`POOL_SNAPSHOTS_WINDOW` on `/pools` page** (`ui-dashboard/src/app/pools/page.tsx`) — 24h safe, 7d crosses the 1000-row cap in ~12 days at current activity. Migrate to paginated fetch when touched next. Warning comment added on the query definition.

#### Performance

3. **Indexer-side `PoolDailyVolume` rollup entity** — [Envio's documented pattern](https://docs.envio.dev/docs/HyperIndex/navigating-hasura). Full-history chart becomes one ~365-row query per pool per year, no pagination, no refresh-time round-trip growth. Previous comment has a schema sketch. Unlocks removal of `fetchAllSnapshotPages` entirely.
4. **Incremental snapshot fetching** — on SWR refresh, fetch only rows newer than the last-seen timestamp, merge into the in-memory array. Cuts per-refresh payload from full-history to `< page_size` in steady state. Less invasive than #3, but still a real refactor around SWR semantics.
5. **Split refresh cadences** — `snapshotsAll` refreshes every 5 min like everything else. Growing payload means refresh cost increases monotonically. Simple lever: separate SWR key with longer `refreshInterval` (15–30 min) specifically for all-history snapshots.
6. **Plotly bundle weight** — `react-plotly.js` is ~2.5MB gzipped. If homepage TTI matters, a lighter lib (Recharts, visx, uPlot) is a measurable win.

#### External

7. **Raise Envio's 1000-row cap / enable `_aggregate` endpoints** on this schema — both are per-docs "intentionally disabled" on Envio Cloud. Worth an ask if the indexer rollup path (#3) is out of reach. If aggregates were enabled, daily-volume could be a single Hasura aggregate query.

#### Minor

8. **`POOL_SNAPSHOTS_CHART` is also missing the `(timestamp, id)` tiebreaker** — same deterministic-ordering concern as `POOL_SNAPSHOTS_ALL` had. Single-pool queries are less likely to have timestamp ties across pools, but worth adding for consistency.
9. **TVL chart doesn't match against a Summary counterpart** — TVL Summary shows current TVL (stock, not a windowed sum) so there's no mismatch to fix today. If we ever add "7d avg TVL" or similar to Summary, the same UTC-day-bucket vs rolling-window semantics pitfall applies to the TVL chart.

---

Most impactful next step if we want to keep investing: **#3 (indexer rollup)** — it cleans up #1, #2, #4, #5, and #6 all at once by making full-history queries cheap enough to not need optimization tricks.

PR #139, opened 2026-04-14
refactor(ui): derive pool detail network from pool id

chapati23:
## Summary

- Drop `?network=` from pool detail URLs — pool ids are namespaced `{chainId}-{address}`, so the chainId is already in the path. Clean URLs like `/pool/143-0xabc...` without the redundant param; callers no longer have to propagate it.
- Preserve `?network=` for non-canonical networks (`devnet`, `celo-mainnet-local`, `celo-sepolia-local`) via a new `buildPoolDetailHref` helper — those share a chainId with a canonical variant, so dropping the param would silently reroute local-dev clicks to the prod indexer.
- Pool-not-found redirect now uses the active network from `useNetwork()` (not the pool id's chainId), so a selector change on pool detail lands the user on the selected chain's `/pools`, not the source chain.

## Implementation

- `networkIdForChainId(chainId)` + `isCanonicalNetwork(id)` in `networks.ts` — canonical-per-chainId map, explicit to avoid drift.
- `NetworkProvider` is now pathname-aware: `?network=` > pathname-derived > `DEFAULT_NETWORK`. `handleNetworkChange` writes the param whenever the selection differs from the implicit (pathname or default) resolution — fixes the silent-no-op where picking Celo on a Monad pool did nothing.
- `buildPoolNotFoundDest(networkId)` signature changed from `(networkParam: string | null)` to take the active `IndexerNetworkId`. Only caller updated.
- `buildPoolDetailUrl` unchanged for in-page URL updates (still passes params through); new `buildPoolDetailHref` is for building fresh listing→detail links.

## Test plan

- [x] `pnpm --filter ui-dashboard test` — 697 pass (16 new NetworkProvider tests, 6 new routing/networks helper tests)
- [x] `pnpm --filter ui-dashboard typecheck` clean
- [x] `trunk check` clean
- [x] Browser (chrome-devtools MCP): `/pool/143-0x...` (no param) → header shows Monad; `/pool/42220-0x...` → Celo; `?network=celo-sepolia` param override wins; selector change on Monad pool writes `?network=celo-mainnet` (fixes silent-no-op); `/pool/42220-...?network=celo-mainnet-local` preserves local context across nav
- [ ] Reviewer: click a pool from homepage cross-chain table, pools page (Celo + Sepolia), global pools — verify clean URLs for canonical networks and preserved param when viewing a local/devnet indexer

### DIFF
--- a/ui-dashboard/src/app/pool/[poolId]/page.tsx
+++ b/ui-dashboard/src/app/pool/[poolId]/page.tsx
@@ -276,15 +276,14 @@ function PoolDetail() {
     searchParams,
   ]);
 
-  // When the pool is not found on the current network (e.g. user switched
-  // networks while viewing a pool), redirect to /pools rather than showing
-  // an error — the pool may simply not exist on the new chain.
-  // Preserve the active ?network= param so the user lands on the correct chain.
+  // Pool not found on the current network → redirect to the active
+  // network's /pools. Using network.id (not the pool id's chainId) honors a
+  // selector change the user may have just made.
   useEffect(() => {
     if (!poolLoading && !poolErr && !pool) {
-      router.replace(buildPoolNotFoundDest(searchParams.get("network")));
+      router.replace(buildPoolNotFoundDest(network.id));
     }
-  }, [pool, poolLoading, poolErr, router, searchParams]);
+  }, [pool, poolLoading, poolErr, router, network.id]);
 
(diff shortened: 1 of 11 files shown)


chapati23:
Addressed in 457ce7b: `buildPoolDetailHref` now preserves `?network=` unless the pool id is namespaced. Added a regression test covering raw `0x…` pool ids on canonical non-default networks.

PR #190, opened 2026-04-22
feat(auth): verify hd + email_verified, enforce allowlist in middleware, 1h JWT

chapati23:
## Summary

Second patch in the red-team remediation series. Closes three Phase 2 findings documented in `.redteam/phase-2-findings.md` (filed alongside this PR).

**Stacked on top of #189** (CSP + HSTS + retire unauth GET). Merge #189 first, then this — the middleware change in this PR relies on the unauth GET path already being gone.

### What changes

- **P2-01 — Google `hd` + `email_verified` in `signIn`.** The existing check was `email.endsWith("@mentolabs.xyz")`. That trusts the email *string*. Google Workspace attackers can game that path in two ways: (a) sign in with a different Workspace whose admin added `mentolabs.xyz` as a secondary domain alias — their OAuth `hd` would be that workspace's primary domain, not ours; (b) pass an unverified profile with a crafted email. Now: `profile.hd === "mentolabs.xyz" && profile.email_verified === true && email.endsWith("@mentolabs.xyz")`. Both `hd` and `email_verified` are Google-server-signed.

- **P2-03 — Middleware enforces the allowlist, not just session presence.** `!!req.auth` let any JWT with a `user.email` through. A forged JWT (AUTH_SECRET leak scenario, since prod and preview share the secret to make `AUTH_REDIRECT_PROXY_URL` work) carrying `email: "attacker@gmail.com"` would reach `/address-book` and the write endpoints. Mirror the domain check at the edge.

- **P2-04 — 1-hour JWT session.** Was 30 days with no revocation list; an offboarded user kept access for a month. Drop `maxAge` to `60 * 60` with `updateAge: 10 * 60` — active users re-mint on every request, idle sessions age out within the hour.

### Not in this PR

- **P2-02 (split AUTH_SECRET between prod and preview)** — rejected. Preview OAuth is bounced through prod via `AUTH_REDIRECT_PROXY_URL`; JWE verification requires the same secret on both envs. Compensating control: enable Vercel Deployment Protection on previews (tracked separately).

## Test plan

- [x] `pnpm exec vitest run` — 1072 passing, includes 10 new cases (7 auth + 3 middleware)
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `trunk check` — clean
- [ ] Preview deployment: sign in with @mentolabs.xyz, verify redirect works (`AUTH_REDIRECT_PROXY_URL` path still functional)
- [ ] Preview deployment: confirm 30-day-old session cookies are invalidated (visible via Application → Cookies; `authjs.session-token` Expires should be ~1h)
- [ ] Preview deployment: hit `/api/address-labels` PUT with no session → 401
- [ ] Preview deployment: idle for >1h, confirm session expires and hitting `/address-book` redirects to `/sign-in`

Rollout note: all currently-signed-in users will be asked to re-authenticate after this deploys (old JWTs carry the old maxAge but middleware will still honor them until they naturally expire — they simply can't extend past their original 30-day window).

---

> [!NOTE]
> **Medium Risk**
> Touches core auth and edge middleware behavior, which can block legitimate users or break API access if the Google profile claims or routing assumptions differ from production.
> 
> **Overview**
> Hardens authentication by tightening the Google `signIn` callback to require `profile.hd === "mentolabs.xyz"`, `email_verified === true`, and an `@mentolabs.xyz` email suffix, and exports `ALLOWED_DOMAIN` for shared enforcement.
> 
> Updates `middleware.ts` to authorize requests based on the session email suffix (not just presence of `req.auth`), and to require auth for **all** `/api/address-labels` methods (keeping `/backup` excluded). Reduces JWT session lifetime from 30 days to **1 hour** (`maxAge`) with a 10-minute `updateAge`, and adds/updates tests to cover the new auth and middleware behaviors (including non-mentolabs and lookalike emails).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3db99f86f7012f69867576ca33e07f94624da96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui-dashboard/src/__tests__/middleware.test.ts
+++ b/ui-dashboard/src/__tests__/middleware.test.ts
@@ -29,12 +29,13 @@ beforeEach(() => {
 
 function makeReq(
   path: string,
-  opts: { method?: string; authenticated?: boolean } = {},
+  opts: { method?: string; authenticated?: boolean; email?: string } = {},
 ): AuthReq {
-  const { method = "GET", authenticated = false } = opts;
+  const { method = "GET", authenticated = false, email } = opts;
   const url = new URL(path, "http://localhost");
+  const resolvedEmail = email ?? (authenticated ? "alice@mentolabs.xyz" : null);
   return {
-    auth: authenticated ? { user: { email: "alice@mentolabs.xyz" } } : null,
+    auth: resolvedEmail ? { user: { email: resolvedEmail } } : null,
     method,
     nextUrl: url,
   } as unknown as AuthReq;
@@ -89,9 +90,12 @@ describe("middleware", () => {
     expect(res!.status).toBe(401);
   });
 
-  it("allows unauthenticated GET /api/address-labels through", () => {
+  it("returns 401 for unauthenticated GET /api/address-labels", () => {
+    // GET is no longer public — the unauth public-label path was retired in
+    // the CSP PR, so middleware now gates every method.
     const res = middlewareCallback(makeReq("/api/address-labels"));
-    expect(res).toBeUndefined();
+    expect(res).toBeDefined();
+    expect(res!.status).toBe(401);
   });
 
   it("returns 401 for unauthenticated /api/address-labels/export", () => {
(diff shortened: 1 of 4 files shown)


chapati23:
@claude review

chapati23:
Thanks for the review — addressing each finding below.

**Finding 1 (duplicated `ALLOWED_DOMAIN`):** Fixed in c3db99f. `ALLOWED_DOMAIN` is now exported from `auth.ts` and imported by `middleware.ts`, so the edge and the sign-in callback can never drift out of sync. All 1072 tests pass.

**Finding 2 (type assertion vs. runtime validation):** Won't fix. Agreed the `as { hd?: string; email_verified?: boolean }` cast is a runtime assumption rather than a compile-time guarantee — but we've bounded the risk: (a) the `account?.provider !== "google"` check above it pins the provider, (b) both checks (`hd !== "mentolabs.xyz"`, `email_verified !== true`) fail *closed* with strict equality, so any missing/misshapen field routes to `return false`, and (c) the middleware domain check (Finding #1) re-enforces the allowlist at the edge, so a drift here wouldn't be load-bearing on its own. Adding zod here would be overkill for the surface area.

**Finding 3 (residual `updateAge` window):** Fixed in c3db99f. Expanded the session comment to document that an offboarded user whose JWT was just re-minted keeps access for up to `updateAge` (10 min) until the next re-mint fails, and that this is the inherent cost of stateless JWTs without server-side revocation.

**Finding 4 (`OPTIONS` preflight returns 401):** Won't fix. The dashboard is internal-only — no CORS surface, no external callers wired up. If we ever do expose an external API on `/api/address-labels/*`, we'll need a dedicated CORS allowlist and preflight handling anyway; that would be a deliberate design change rather than a side effect of loosening middleware today.

PR #218, opened 2026-04-24
fix(bridge): correct IAM + logging config so bridge deploy reaches Cloud Run

chapati23:
## Summary

Follow-up to PR #216, which grafted `roles/serviceusage.serviceUsageConsumer` onto the deploy SA based on the gcloud CLI's misleading error message. That role didn't actually unblock the deploy — verified by applying it live and retrying.

Root cause isolation via `gcloud builds submit --verbosity=debug` exposed the real errors:

1. `gcloud builds submit` runs a **project-scoped** `storage.buckets.list` probe to resolve the default staging bucket. Bucket-scoped storage IAM doesn't satisfy it. Need `roles/storage.admin` at project level.
2. The CLI then polls build logs, and the default Cloud Build log destination is a GCS bucket that requires project Viewer/Owner to stream. Pinning `options.logging: CLOUD_LOGGING_ONLY` keeps logs in Cloud Logging so `roles/logging.viewer` is sufficient.

### `terraform/main.tf`
- Swap `roles/serviceusage.serviceUsageConsumer` → `roles/storage.admin` + `roles/logging.viewer`
- Drop the bucket-scoped `google_storage_bucket_iam_member` (redundant now that project-level `storage.admin` covers it)
- Expand comment with the actual root cause so a future reader doesn't chase the CLI's red herring

### `cloudbuild.yaml`
- Add `options.logging: CLOUD_LOGGING_ONLY` so build logs go to Cloud Logging, not the default GCS log bucket. Lets `roles/logging.viewer` cover the streaming side without needing project Viewer.

## Verification

Applied live, reproduced end-to-end locally with impersonation:

\`\`\`
$ gcloud builds submit --impersonate-service-account=metrics-bridge-deployer@mento-monitoring.iam.gserviceaccount.com ...
…
STATUS: SUCCESS
\`\`\`

Image pushed to `europe-west1-docker.pkg.dev/mento-monitoring/metrics-bridge/metrics-bridge:test-local-*`.

## Test plan

- [x] Local impersonated `gcloud builds submit` completes successfully
- [ ] After merge: re-dispatch `metrics-bridge.yml` and confirm `Build and push image` + `Deploy to Cloud Run` both green
- [ ] Cloud Run revision `metrics-bridge` picks up the new image


chapati23:
@claude review pls

chapati23:
@claude[bot] thanks for the review ✅

The `storage.admin` tightening note is on the radar — the inline Terraform comment now spells out the single-tenant rationale and the "tighten to a custom role if this project ever hosts sensitive GCS data" follow-up. Not fixing in this PR because (a) `mento-monitoring` only hosts metrics-bridge + its cloudbuild artifacts (no sensitive buckets to widen-access to), (b) a custom role adds maintenance surface disproportionate to the marginal security gain here, and (c) the real blast-radius amplifier is the WIF `principalSet` at terraform/main.tf:523 (any workflow in the repo, not just main) which pre-dates this PR and is worth a focused follow-up PR on its own.

Shipping as-is.

PR #239, opened 2026-04-28
fix(metrics-bridge): handle REBALANCE_PROBE_EVERY_N_POLLS=1 correctly

chapati23:
## Summary

`metrics-bridge/src/poller.ts` had a silent foot-gun in the rebalance-probe cadence:

```ts
if (pollCycle % REBALANCE_PROBE_EVERY_N_POLLS !== 1) return;
```

For `N >= 2` this works fine. For `N = 1`, `pollCycle % 1` is **always** `0`, so `!== 1` is **always** `true` and the early-return **always** fires — `EVERY_N=1` (intended as "probe every poll") silently disables the probe entirely. Default config ships `N=5`, so this is a future-config foot-gun rather than a current bug, but an operator dialling cadence down to "every poll" would get the opposite of what they asked for.

Flagged by Codex P2 + Cursor on PR #235; tracked in `docs/BACKLOG.md`.

## Fix

Flipped to a **0-indexed cycle counter** with the increment moved to AFTER the probe check on successful polls. New predicate: `(pollCycle % N) === 0`.

- Cold-start invariant preserved — cycle 0 always fires (any N), so an operator restarting the bridge mid-breach doesn't wait N×30s for the reason annotation.
- Failed Hasura polls still don't advance the counter (the `pollCycle += 1` line throws-past on error), so the probe still attaches to the next successful cycle.
- Predicate is now well-defined for all `N >= 1`. No special case.

Considered the alternatives (option (b) `(cycle % N) !== (1 % N)` and option (a) explicit `N > 1` guard); the 0-indexed counter is the cleanest read at the call site. Reasoning is documented in the in-file comment block.

## Tests locking the new behaviour (4 new in a `rebalance probe cadence` describe)

- **`EVERY_N=1` regression** — probe fires on every poll (5/5 over 5 polls). Comment cross-references the BACKLOG foot-gun so future readers see why the test exists.
- **`EVERY_N=2`** — probe fires on cycles 0/2/4 (3/6 over 6 polls).
- **`EVERY_N=5` (default)** — probe fires on cycles 0/5 (2/6 over 6 polls). Same observable behaviour as the existing default-cadence test.
- **`EVERY_N=0` boundary** — config parser already rejects 0 via the `>= 1` guard in `metrics-bridge/src/config.ts`; new test confirms the constant falls back to 5 and the cadence behaves accordingly. So an operator typo (`EVERY_N=0`) does NOT silently disable the probe either.

The four new tests use `vi.stubEnv` + `vi.resetModules` + dynamic import because `REBALANCE_PROBE_EVERY_N_POLLS` is captured as a const at module-load. Existing cold-start + failed-poll-doesn't-advance + per-error-path tests stay green.

## BACKLOG

Moved "Rebalance probe: handle `REBALANCE_PROBE_EVERY_N_POLLS=1`" from Tech Debt into the v3 Alerting Done section.

## Test plan

- [x] `pnpm --filter @mento-protocol/metrics-bridge test` — 153 tests pass (4 new cadence tests; previously 149)
- [x] `pnpm --filter @mento-protocol/metrics-bridge typecheck`
- [x] `pnpm --filter @mento-protocol/metrics-bridge lint`
- [x] `pnpm --filter @mento-protocol/metrics-bridge build`
- [x] `pnpm exec trunk fmt && pnpm exec trunk check --all` — clean

## Files

- `metrics-bridge/src/poller.ts` — predicate flip + cycle increment moved past probe check
- `metrics-bridge/test/poller.test.ts` — 4 new cadence tests + comment refresh on existing N=5 test
- `docs/BACKLOG.md` — entry moved Tech Debt → Done

---

> [!NOTE]
> **Low Risk**
> Low risk: a small, well-tested change to rebalance-probe scheduling logic and associated tests; no auth/security or data model changes.
> 
> **Overview**
> Fixes a cadence edge case where setting `REBALANCE_PROBE_EVERY_N_POLLS=1` would silently disable the rebalance probe by switching the poll cycle logic to a 0-indexed counter and running the probe when `(cycle % N) === 0` (with the cycle increment moved to after the probe check).
> 
> Adds regression tests that re-import `poller` under different env values to lock probe behavior for `N=1`, `N=2`, `N=5`, and to confirm invalid `N=0` falls back to the default cadence; also updates `docs/BACKLOG.md` to mark the foot-gun as done.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ada0e4025a20c7f6678bdd432f6e3383ea5fb91f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/BACKLOG.md
+++ b/docs/BACKLOG.md
@@ -1,6 +1,6 @@
 # Monitoring Monorepo — Task Backlog
 
-Last updated: 2026-04-24
+Last updated: 2026-04-28
 
 ## Next — Indexer: CDP strategy entity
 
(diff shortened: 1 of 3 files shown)


chapati23:
@claude review

chapati23:
cursor review

PR #240, opened 2026-04-28
fix(metrics-bridge): re-entrancy guard for overlapping probe cycles

chapati23:
## Summary

- Adds a module-scope `probeInProgress` mutex around `runRebalanceProbes` in `metrics-bridge/src/rebalance-probe.ts`. A second call while a previous cycle is still in flight short-circuits with a single warn-once `[REBALANCE_PROBE_REENTRY]` log line instead of resetting the gauge and double-writing labels.
- Production today doesn't trigger this — `metrics-bridge/src/poller.ts` awaits each `runRebalanceProbes()` call before scheduling the next loop, so cycles can't overlap. The guard is a safety rail for future async restructures (e.g. the AbortController fast-cancel path tracked separately in BACKLOG, or a parallel-timer caller).
- Flips the existing limitation-locking test (`metrics-bridge/test/rebalance-probe.test.ts`, previously titled "documents the known limitation: late cycle-N-1 results stomp cycle-N gauge state") to assert the new correctness: only one probe runs, gauge state is preserved, no double-execution, and the warn log fires exactly once. Renamed to `runRebalanceProbes — re-entrancy guard for overlapping cycles` so git blame on the rationale is preserved.
- Removes the `Rebalance probe: re-entrancy guard for overlapping cycles` entry from `docs/BACKLOG.md`.

## Why now

The AbortController PR (`metrics-bridge/probe-abort-controller`, parallel work) could change probe-completion timing semantics enough that overlapping cycles become reachable. Adding the guard ahead of time keeps the surface area narrow — that PR can land without needing a follow-up safety patch. The `try { ... } finally { probeInProgress = false }` shape ensures the flag releases on probe throws, so a transient error can't permanently disable the probe.

## Test plan

- [x] `pnpm --filter @mento-protocol/metrics-bridge test` — 149 passed (28 unchanged + 1 flipped)
- [x] `pnpm --filter @mento-protocol/metrics-bridge typecheck`
- [x] `pnpm --filter @mento-protocol/metrics-bridge lint`
- [x] `pnpm --filter @mento-protocol/metrics-bridge build`
- [x] `pnpm exec trunk fmt && pnpm exec trunk check --all` — no issues
- [ ] Existing `runRebalanceProbes` tests (timeout race, gauge writes, reset semantics, reserve enrichment) all still pass — guard is invisible to non-overlapping callers

## Notes

The new test mocks the cycle-1 probe with a manually-resolvable promise to pin the runner mid-flight, kicks off cycle 2 (which short-circuits), then resolves cycle 1 and asserts its single labelset lands as expected. The mutex check is purely synchronous so cycle 2's no-op happens before any microtask yields back to cycle 1.

---

> [!NOTE]
> **Medium Risk**
> Changes the rebalance-probe runner’s execution semantics by skipping concurrent runs, which could affect alert-annotation metrics if overlap occurs (though production is currently serialized); behavior is covered by new unit tests.
> 
> **Overview**
> Prevents overlapping `runRebalanceProbes()` cycles from racing by adding a module-scope re-entrancy guard: if a cycle is already running, subsequent calls short-circuit without resetting or writing any gauges and emit a warn-once `[REBALANCE_PROBE_REENTRY]` log.
> 
> Updates the Vitest suite to assert the new skip-on-busy behavior (including *warn-once-per-busy-window* and mutex release on probe throw) and adds a test-only reset hook for the module-scope flag. Removes the corresponding re-entrancy tech-debt item from `docs/BACKLOG.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3951cfbdd0b2363a0c67733c093a1ee486a0e5dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chapati23:
@claude review

chapati23:
cursor review

chapati23:
Fixed in 0f5294e — added a negative assertion after the second cycle in the "releases the mutex when a probe throws" test, confirming no `[REBALANCE_PROBE_REENTRY]` warn is emitted on a throw-then-recover sequence.

PR #413, opened 2026-05-13
perf(indexer): preload trading limit reads

chapati23:
## Summary

- Preload `tradingLimitsEffect` for both FPMM swap tokens during Envio's preload phase.
- Keep the existing processing-phase writes unchanged: TradingLimit rows, Pool limit pressure/status, SwapEvent, snapshots, and leaderboard rollups still persist only in the ordered processing pass.
- Split the Pool preload helper so callers can reuse the existing Pool/open-breach warm-up while also reading the warmed Pool row for dependent effect preloads.

### What this PR changes

This changes when the per-swap trading-limit RPC effects are requested, not what gets written. The `FPMM.Swap` handler now requests those effect reads during preload so Envio can batch and parallelize them across the event batch, then reuses the effect result during the processing pass.

### Invariants

- `FPMM.Swap` still performs entity writes only during the processing pass.
- Trading limits remain block-scoped and `cache: false`; this PR does not make historical trading-limit state durable-cached.
- Missing Pool/token metadata in preload degrades to the previous behavior: the processing pass still performs the authoritative update path.
- The existing Pool and open-breach preload warm-up remains intact.

### Degraded behavior

RPC failure behavior is unchanged. If `tradingLimitsEffect` returns `null`, the processing pass skips the missing limit row and retries on a later swap, as before. If preload cannot find a complete Pool row, it only skips the optimization and the ordered processing pass still runs the existing logic.

### Intentional non-goals

- No schema, GraphQL, dashboard, or entity-shape changes.
- No change to persistent effect caching for trading limits.
- No local derivation of trading-limit state from swap events.

## Test plan

- `pnpm indexer:testnet:codegen`
- `pnpm indexer:codegen`
- `CI=true pnpm install --frozen-lockfile`
- `./tools/trunk check --all`
- `pnpm --filter @mento-protocol/indexer-envio lint`
- `pnpm --filter @mento-protocol/indexer-envio typecheck`
- `pnpm --filter @mento-protocol/indexer-envio test`

---

> [!NOTE]
> **Medium Risk**
> Changes `FPMM.Swap` control flow across Envio preload/processing passes and alters when `tradingLimitsEffect` is invoked, which could impact indexing behavior/performance if preload data is incomplete or effect key reuse differs from expectations.
> 
> **Overview**
> During Envio's preload pass, `FPMM.Swap` now warms the Pool cache and pre-requests `tradingLimitsEffect` for both swap tokens so RPC reads can be batched/parallelized, while keeping all entity writes in the ordered processing pass.
> 
> To support dependent preloads, the pool preload helper is split: `maybePreloadPool` still early-returns for preload-only warmups, and new `preloadPoolCache` returns the warmed `Pool` (and open breach row) to callers that need its fields before issuing effects; tests are updated to cover the new helper behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c61b9f4af191cbafb7898fc3f1ebac8ce6ed39f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/indexer-envio/src/handlers/fpmm.ts
+++ b/indexer-envio/src/handlers/fpmm.ts
@@ -12,7 +12,7 @@ import {
   tradingLimitId,
 } from "../tradingLimits.js";
 import { tradingLimitsEffect } from "../rpc/effects.js";
-import { maybePreloadPool, upsertPool, upsertSnapshot } from "../pool.js";
+import { preloadPoolCache, upsertPool, upsertSnapshot } from "../pool.js";
 import { buildSwapTraderFields } from "../swap.js";
 import { applyLeaderboardSnapshots } from "../leaderboardSnapshots.js";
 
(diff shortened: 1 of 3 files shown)


chapati23:
@claude review

chapati23:
Fixed in c61b9f4 — added rationale comments documenting that the processing pass reuses the preload effect keys/results, and that preloadPoolCache is safe to call from either pass.

PR #417, opened 2026-05-13
perf(indexer): group swap trading limit effects

chapati23:
## Summary
- Add a pool-level `poolTradingLimits` effect that reads both FPMM token trading-limit states under one Envio effect envelope.
- Update `FPMM.Swap` preload and processing paths to use the grouped effect, keeping authoritative at-block reads while reducing hot-path effect scheduling from two effect calls per swap to one.

## What this PR changes
The swap handler still persists the same `TradingLimit` rows and `Pool.limitStatus` fields. The only behavior change is the effect envelope used to fetch token0/token1 trading limits during replay.

## Invariants
- Trading-limit state remains block-scoped and `cache: false`; no persisted Effect API cache is used for these values.
- Both token limit states are fetched at the swap block and each `TradingLimit` row is written only from its matching token result.
- Partial RPC failure behavior is unchanged: if one token read fails, the other token row can still update and the pool limit status reflects available data only.

## Degraded behavior
- If both token trading-limit reads fail, the grouped effect returns `null` and the swap skips trading-limit writes as before.
- If exactly one token read fails, the existing partial-update warning path remains in place.
- No schema, GraphQL, dashboard, or generated type changes are required.

## Intentional non-goals
- Does not locally derive TradingLimitsV2 state from swap amounts; that would risk fee-history and contiguity drift during replay.
- Does not change hosted cache settings or deploy a new indexer. We will deploy after merge to measure full replay time.

## Test plan
- `pnpm --filter @mento-protocol/indexer-envio typecheck`
- `pnpm --filter @mento-protocol/indexer-envio test test/tradingLimits.test.ts test/swap-reserves.test.ts test/effects.test.ts`
- `pnpm agent:quality-gate --run`
- push hook reran the mapped gate successfully: codegen, frozen install, Trunk, lint, typecheck, strict typecheck, full indexer test suite (`631 passed`)

Checklist: reviewed `docs/pr-checklists/stateful-data-ui.md`; this change touches an event handler/effect path only, with no entity/schema/query shape change.

---

> [!NOTE]
> **Medium Risk**
> Touches the FPMM swap hot path and changes how per-swap trading-limit RPC reads are scheduled/deduped, which could impact replay performance or edge-case failure handling. No schema changes, but correctness depends on the new grouped effect returning token-specific results consistently.
> 
> **Overview**
> **Reduces per-swap effect overhead for FPMM trading limits** by introducing `poolTradingLimitsEffect`, which fetches token0/token1 limit state together under a single Envio Effect API call.
> 
> Updates the `FPMM.Swap` preload and processing paths to use this grouped effect (instead of two `tradingLimitsEffect` calls), while still writing the same `TradingLimit` entities and computing `Pool` `limitStatus` from whatever token data is available.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd74426a5a2ffcce677ab58cc1381cae3d22e0e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/indexer-envio/src/handlers/fpmm.ts
+++ b/indexer-envio/src/handlers/fpmm.ts
@@ -11,7 +11,7 @@ import {
   computeLimitStatus,
   tradingLimitId,
 } from "../tradingLimits.js";
-import { tradingLimitsEffect } from "../rpc/effects.js";
+import { poolTradingLimitsEffect } from "../rpc/effects.js";
 import { preloadPoolCache, upsertPool, upsertSnapshot } from "../pool.js";
 import { buildSwapTraderFields } from "../swap.js";
 import { applyLeaderboardSnapshots } from "../leaderboardSnapshots.js";
@@ -37,21 +37,14 @@ indexer.onEvent(
       ) {
         // Keep processing writes out of preload, but let Envio batch and
         // parallelize the per-swap trading-limit RPC reads. The processing
-        // pass presents the same effect keys and reuses these preload results.
-        await Promise.all([
-          context.effect(tradingLimitsEffect, {
-            chainId: event.chainId,
-            poolAddress: event.srcAddress,
-            token: pool.token0,
-            blockNumber,
-          }),
-          context.effect(tradingLimitsEffect, {
-            chainId: event.chainId,
-            poolAddress: event.srcAddress,
-            token: pool.token1,
-            blockNumber,
-          }),
-        ]);
+        // pass presents the same effect key and reuses these preload results.
+        await context.effect(poolTradingLimitsEffect, {
+          chainId: event.chainId,
+          poolAddress: event.srcAddress,
+          token0: pool.token0,
+          token1: pool.token1,
+          blockNumber,
+        });
       }
       return;
     }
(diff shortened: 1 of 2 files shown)


chapati23:
@claude review

chapati23:
Addressing Claude's review notes for this measurement PR:

- Keeping tradingLimitsEffect as-is for now because this PR is intentionally scoped to the swap hot path. If the grouped path proves out after hosted replay, we can prune the legacy single-token effect in a follow-up cleanup.
- Keeping poolTradingLimitsEffect at rateLimit 200 per second deliberately for the deploy experiment. Lowering it to 100 would preserve the old theoretical RPC-call ceiling, but it would also mask whether the Envio effect scheduling overhead was the bottleneck we are trying to measure. We should watch provider throttling/fallback logs during the new indexer sync.
- Not adding extra unit coverage in this PR because the partial-null behavior is still driven by the existing fetchTradingLimits contract and covered by the integration path we already ran. If we keep this shape after the hosted measurement, focused effect tests are a reasonable follow-up.

No code changes from the review notes for this PR.

PR #472, opened 2026-05-19
fix(cdps): correct GBPm SystemParams + retire devnet remnants

chapati23:
## Summary

- **Correctness fix**: GBPm `systemParams` in `indexer-envio/config/liquity.json` was pointing at `0xddd2de5db7e81e22553cc68ea6125b8a817ac8ea` — an address with **zero bytecode** on Celo. Verified via `eth_getCode`. As a result, `getOrLoadSystemParams` failed silently for GBPm and our defensive fallback classified every non-zero-debt GBPm trove as `zombie`, regardless of whether it was actually below `MIN_DEBT`. The real address is `0x064d8bcc79711cf51df7ca0a7fe531a271cd74e9` (`SystemParamsv300GBPm` per `@mento-protocol/contracts/contracts.json`), verified live with MIN_DEBT=1000 / MCR=110% / CCR=135%.
- **Root cause**: the wrong address was copy-pasted from `ui-dashboard/src/lib/addresses.json` — a devnet-only address book that was never kept in sync with mainnet. The devnet itself was torn down (hence the dead address).
- **Architectural cleanup**: `@mento-protocol/contracts/contracts.json` is the canonical source for all protocol addresses across all networks. Double-tracking via a hand-vendored devnet file is what caused the bug. This PR also deletes the orphan devnet machinery: the address book, the sync script that populates it (`sync-contract-config.mjs`), its output (`contracts.celo.v3.json`), the `devnet` IndexerNetworkId + `NETWORKS.devnet` entry, every consumer's handling of devnet, and the related env-var docs. The `celo-mainnet-local` and `celo-sepolia-local` network entries stay — they're functionally distinct (real-RPC local dev, not torn-down anvil).

After this merges and the indexer resyncs, GBPm troves currently labeled `zombie` (real debts of 60k–150k, well above `MIN_DEBT=1000`) will correctly classify as `active`.

## What "zombie" actually means in Liquity v2

For the avoidance of future confusion: a Liquity v2 zombie is a trove whose recorded debt fell *below* `MIN_DEBT` — almost always because a redemption swept it down to dust. The trove still holds debt + collateral and counts toward system debt, but the owner can't adjust/close until they top it back up (`adjustZombieTrove`). It's a dust-management mechanism. Our indexer's defensive fallback co-opted the label as a "params not loaded" signal, which silently created the bug surface.

## Test plan

- [x] Both packages typecheck clean
- [x] 2204 dashboard + 695 indexer tests passing (added one new test: the hasura proxy returns 404 for the retired `devnet` network)
- [x] `trunk fmt` + `trunk check` clean
- [x] `pnpm lint` clean in both packages (1 stale ESLint baseline entry pruned in `ui-dashboard` — `resolveLocalHasuraConfig` complexity dropped from 16 → 15 after removing the devnet branch)
- [x] Cross-checked all 30 protocol-contract addresses in `indexer-envio/config/liquity.json` against `@mento-protocol/contracts/contracts.json` — every one matches after this fix
- [x] Probed the new GBPm `systemParams` address on-chain: MCR=110%, CCR=135%, MIN_DEBT=1000 (matches CHFm)
- [ ] Post-merge: deploy indexer + full resync (the prior resync from PR #470 produced a deployment that's not promoted; this fix needs to land before that gets promoted, so the new resync produces correct GBPm classification)

## Commits

1. `cc2dc063` — the one-line config fix (correctness)
2. `0ce80be7` — devnet remnants kill (architectural cleanup so this can't happen again)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes on-chain address configuration used by the indexer for Liquity/CDP classification and removes a network variant (`devnet`), which could affect local tooling and routing if any dependency still expects it.
> 
> **Overview**
> **Fixes a misconfigured Liquity GBPm `systemParams` address** in `indexer-envio/config/liquity.json`, correcting downstream CDP/trove classification that depended on reading those params.
> 
> **Retires the old Celo devnet plumbing** across the repo: deletes the devnet address book/artifact sync outputs and script, removes `devnet` from dashboard network definitions and the local Hasura proxy/env var docs, and updates tests/baselines to assert `devnet` is now unsupported (404) while keeping `celo-*-local` variants intact.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ce80be7b6b58710d3e1f93f722d72e999f85689. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chapati23:
@claude review

PR #500, opened 2026-05-21
fix(claude-hooks): SessionEnd resilient to deleted worktrees

chapati23:
## Summary

- Guard the project `SessionEnd` hook on script existence (`if [ -f X ]; then bash X; fi`) so it silently no-ops when a worktree is removed while a Claude Code session inside it is still alive, instead of failing with `bash: …: No such file or directory` on every such session exit.
- Update `scripts/check-agent-context.mjs` validator regex to accept both the original direct-invocation form and the new presence-guarded form, so `pnpm agent:context-check` stays green.

## Test plan

- [x] `pnpm agent:context-check` → exit 0
- [x] Unset / missing / spaces in `CLAUDE_PROJECT_DIR` → hook exits 0 silently
- [x] Hook script exists + exits 0 → hook exits 0
- [x] Hook script exists + exits non-zero → hook propagates that exit code (failure surfaces; not swallowed)
- [x] `trunk fmt` + `trunk check` clean
- [x] Pre-push gate: `./tools/trunk check` + `pnpm agent:context-check` + `pnpm lint:scripts` all pass

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to developer hook configuration and its regex validator, with no impact on application runtime logic.
> 
> **Overview**
> Makes the Claude `SessionEnd` hook resilient to deleted worktrees by guarding execution with `if [ -f ... ]; then bash ...; fi`, avoiding repeated “No such file” failures when the hook script path no longer exists.
> 
> Updates `scripts/check-agent-context.mjs` to validate both the original direct `bash "${CLAUDE_PROJECT_DIR}/..."` form and the new presence-guarded form so `pnpm agent:context-check` continues to pass.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52a403036daac04d0687bdc2143bb379b3714225. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\""
+            "command": "[ -f \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\" ] && bash \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\" || true"
           }
         ]
       }
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "[ -f \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\" ] && bash \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\" || true"
+            "command": "if [ -f \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\" ]; then bash \"${CLAUDE_PROJECT_DIR}/scripts/agent-session-end-hook.sh\"; fi"
           }
         ]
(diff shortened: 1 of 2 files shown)


chapati23:
Two non-blocking notes from the auto-review:

1. **Validator error message wording** (`scripts/check-agent-context.mjs:297`). The `fail(...)` text describes only the direct-invocation form. Won't fix — the validator only fires when the command shape is genuinely wrong, and anyone debugging will read the regex (lines 260-271) as the source of truth, not the error string. Updating the message to enumerate both accepted forms adds maintenance burden each time the set of accepted shapes changes.

2. **Regex tolerates mismatched quote types** (`scripts/check-agent-context.mjs:260-261`). `scriptPath` is `["']\${…}["']`, so opening/closing quotes don't have to match. Won't fix — `.claude/settings.json` is machine-written, never hand-rolled, and the validator's contract here is path correctness, not quote hygiene. The auto-reviewer noted the same self-deferral.

PR #821, opened 2026-06-09
ci: ARM runner compat test (throwaway — do not merge)

chapati23:
## The Problem

- We hypothesize ~37.5% cheaper CI minutes by moving Blacksmith jobs from x64 ($0.004/min base) to ARM (`blacksmith-{2,4}vcpu-ubuntu-2404-arm`, $0.0025/min base), but per-job ARM compatibility has never been tested empirically in this repo.
- Two risks can only be settled by running the jobs: (1) the ui job's Playwright visual-snapshot baselines were rendered on x64 Chromium and carry no arch suffix; (2) the required `Code Quality` check downloads the trunk CLI + its hermetic tools per-platform.

## The Solution

- **Throwaway draft PR — never merges.** It flips every Blacksmith job (except `lighthouse.yml`, see below) to the same-vCPU ARM label so all PR-triggered workflows execute once on ARM, plus a trivial `ui-dashboard/src` comment so `size-limit`, `ui`, and the Vercel preview (needed by Lighthouse) also trigger.
- Non-PR read-only workflows (`mutation-testing`, `terraform-drift`, `integration-probes`, `infra`) are exercised via `workflow_dispatch` from this branch.
- The real migration PR follows after these results + a review pass; jobs that fail here stay on x64 there.

## Details

- `lighthouse.yml` intentionally stays on x64: lhci launches Chrome via chrome-launcher and Google ships no Chrome for linux-arm64.
- Playwright Chromium cache keys gain `${{ runner.arch }}` in both `ci.yml` and `lighthouse.yml` — they share a key namespace but now run on different architectures; a cross-arch restore would pass version validation and die with an exec-format error.
- Envio codegen cache key gains `${{ runner.arch }}` as well.
- Both actionlint `self-hosted-runner` allow-lists learn the `-arm` labels.
- Toolchain arm64 availability pre-verified: trunk CLI 1.25.0 tarball, `envio-linux-arm64`, foundry `linux_arm64`, Playwright Chromium arm64, esbuild/sharp/lightningcss arm64 variants in the lockfile, codecov CLI linux-arm64, and linux_arm64 builds for all 14 Terraform providers across the 4 stacks (lock files carry `zh:` hashes, so cross-platform `terraform init` needs no lockfile edits).

## Validation

- This PR **is** the validation run — results feed the migration decision matrix per job.
- `trunk fmt` + `trunk check` clean on all changed files locally.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -2,3 +2,5 @@ self-hosted-runner:
   labels:
     - blacksmith-4vcpu-ubuntu-2404
     - blacksmith-2vcpu-ubuntu-2404
+    - blacksmith-4vcpu-ubuntu-2404-arm
+    - blacksmith-2vcpu-ubuntu-2404-arm
--- a/.github/workflows/aegis-app-engine.yml
+++ b/.github/workflows/aegis-app-engine.yml
@@ -37,7 +37,7 @@ jobs:
     environment:
       name: production
       url: https://mento-monitoring.uc.r.appspot.com
-    runs-on: blacksmith-4vcpu-ubuntu-2404
+    runs-on: blacksmith-4vcpu-ubuntu-2404-arm
     timeout-minutes: 15
     permissions:
       contents: read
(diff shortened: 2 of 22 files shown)


chapati23:
Compat testing complete — two full ARM sweeps green (the one Code Quality failure was trunk-action cross-arch cache poisoning, root-caused and fixed via arch-keyed cache-key). Durations captured; the scoped migration PR follows with the evidence table. Closing this throwaway branch.

PR #1271, opened 2026-07-15
fix(dashboard): validate SSR GraphQL fallbacks

### LINKED ISSUE #1215: Validate SSR-prefetched fallback payloads with the same Zod schemas the client uses
## Problem

SSR-prefetch paths hand raw GraphQL responses to SWR as `fallbackData`, bypassing `useGQL`'s Zod schema gate: `lib/volume-ssr.ts` (requestOptional returns raw `client.request<T>()`) and the pre-existing `lib/pool-detail-ssr.ts` do exactly the same. A malformed hosted-Hasura response (schema drift/lag) reaching the server render is typed but unvalidated — worst case it throws during the server render or paints wrong shapes for one poll interval until the client's schema-gated revalidation self-heals.

## Solution ideas

Run the same Zod schemas (`lib/queries/volume-schemas.ts`, pool-detail equivalents) over the server-fetched responses inside the SSR modules; on parse failure return `undefined` (the documented fail-open path) instead of forwarding unvalidated data. Keep the schemas shared so server and client can't drift.

## Done means

Both `volume-ssr.ts` and `pool-detail-ssr.ts` validate every prefetched response before it crosses to the client, with tests for the parse-failure → undefined path.

chapati23:
## The Problem

- Volume and pool-detail SSR fallbacks trusted unvalidated GraphQL responses even though matching client subscriptions already used runtime schemas.
- A malformed optional response could cross the server/client boundary or make an entire fallback fail instead of degrading only that extension.

## The Solution

- Validate every prefetched volume and pool-detail response with the exact shared Zod schema used by its client useGQL subscriber.
- Drop malformed optional extensions independently and return no overall fallback when a required base response is malformed, preserving the existing fail-open client revalidation path.

## Verification

- pnpm agent:quality-gate --run
- CI=true pnpm agent:autoreview
- Focused SSR/schema tests: 137 tests
- Full dashboard coverage: 3,825 tests
- React Doctor score 100
- Independent semantic review: no actionable findings
- Direct Chrome: /volume completed without its loading boundary and without console errors.
- Direct Chrome: canonical Celo pool rendered its validated header, oracle, reserve, TVL, and volume data with no loading boundary or console errors.

## Stateful data/UI checklist

- Invariant: a server fallback and its matching client subscriber accept the same response shape through the same schema object.
- Degraded behavior: malformed optional extensions disappear independently; malformed required base data yields undefined so the existing client fetch remains authoritative.
- Non-goals: change queries, polling cadence, cache keys, timeouts, or user-facing error policy.

Closes #1215

---

> [!NOTE]
> **Medium Risk**
> Touches first-paint SSR fallbacks for pool detail and volume hero data; wrong schema strictness could strip valid fallbacks, but behavior intentionally fails open to client revalidation without changing queries or cache keys.
> 
> **Overview**
> **Pool-detail and volume SSR prefetches now run every GraphQL response through the same Zod schemas their matching `useGQL` subscribers use**, so server and client agree on shape instead of trusting raw Hasura JSON.
> 
> `pool-detail-ssr.ts` and `volume-ssr.ts` **`safeParse` each optional fetch**: a bad **required** base payload drops the whole fallback (`undefined`); a bad **optional** extension is omitted on its own (e.g. malformed thresholds but valid v2 exchange). Malformed v2 exchange also skips the dependent 24h volume request.
> 
> **New schemas** in `pool-detail-schemas.ts` cover thresholds, VP extensions, v2 exchange, broker 24h snapshots, and breaker config; volume SSR uses existing `volume-schemas`. **Client hooks and header/breaker/oracle components** pass those schema objects into `useGQL` alongside existing `fallbackData` and timeouts. **Tests** assert schema wiring and SSR degradation for wrong-shaped responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffb796d114f4ffcd3d7fdf381f2ed2f6b2caeb60. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui-dashboard/src/app/pool/[poolId]/__tests__/page.test.tsx
+++ b/ui-dashboard/src/app/pool/[poolId]/__tests__/page.test.tsx
@@ -3,6 +3,11 @@ import { renderToStaticMarkup } from "react-dom/server";
 import type { ReactNode } from "react";
 import type { PoolDetailInitialData } from "@/lib/pool-detail-initial-data";
 import type { PoolBreakerConfigResponse } from "@/lib/queries/config";
+import {
+  BrokerExchangeDailySnapshots24hSchema,
+  PoolBreakerConfigSchema,
+  PoolV2ExchangeSchema,
+} from "@/lib/queries/pool-detail-schemas";
 import type { Pool } from "@/lib/types";
 
 const mockUseGQL = vi.fn();
@@ -654,12 +659,14 @@ describe("Pool detail LPs tab", () => {
 
     expect(findUseGqlCall("PoolV2Exchange")?.[3]).toMatchObject({
       fallbackData: initialData.v2Exchange,
+      schema: PoolV2ExchangeSchema,
     });
     expect(
       findUseGqlCall("BrokerExchangeDailySnapshots24h")?.[3],
     ).toMatchObject({
       timeoutMs: 5000,
       fallbackData: initialData.brokerExchange24h,
+      schema: BrokerExchangeDailySnapshots24hSchema,
     });
     expect(html).toContain("ConstantSum");
     expect(html).not.toContain("v2 exchange config unavailable");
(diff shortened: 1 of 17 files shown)


chapati23:
Fixed the cosmetic doc-comment wrap in a6318334.

PR #1385, opened 2026-07-21
fix(alerts): harden QuickNode webhook state reads

chapati23:
## The Problem

- Alerts Infra plans could fail during QuickNode refresh with the unhelpful `giving up after 1 attempt(s)` error because the REST provider issued concurrent, single-attempt reads.
- The webhook recovery script could mistake nested response IDs for Terraform's canonical resource ID and could offer destructive state removal after an inconclusive network failure.

## The Solution

- Serialize QuickNode provider requests and make webhook state recovery fail closed, with bounded retries and exact top-level Terraform ID parsing.

## Details

- Sets the QuickNode REST provider to one request at a time without enabling unsafe global retries for POST/DELETE operations.
- Shares strict resource-ID extraction between the diagnostic script and replacement provisioner.
- Prevents state removal after transport errors, rate limits, server failures, or ambiguous Terraform output.
- Adds a real-shell regression suite covering nested provider IDs, transient 503 responses, and duplicate top-level IDs, and wires it into Alerts CI and the agent quality gate.
- Updates the Alerts runbooks, package context, and generated docs catalog.
- Investigated failure: https://github.com/mento-protocol/monitoring-monorepo/actions/runs/29829834112

## Validation

- `pnpm agent:quality-gate --run --parallel 1`
- `pnpm agent:quality-gate:test`
- `pnpm agent:context-check`
- `pnpm agent:autoreview` plus two fresh-context semantic reviews (final result: clean)
- `CI=true AGENT_PREWARM_PARALLELISM=1 pnpm agent:prewarm --base origin/main`
- Authenticated read-only `terraform plan` against the live Alerts backend completed successfully (`17 add, 1 change, 17 destroy`); no apply was performed.
- Live webhook reads confirmed the Celo, Ethereum, and Polygon resources remain active.

## Checklist

- [x] Nothing was knowingly deferred.
- [x] No architectural decision was introduced; this hardens the existing QuickNode integration.

---

> [!NOTE]
> **Medium Risk**
> Touches alerts infra Terraform and state-repair scripts that can remove or replace live QuickNode webhooks; changes are defensive but operate on production webhook delivery paths.
> 
> **Overview**
> Hardens **QuickNode webhook Terraform state handling** so plans and repair flows stop mis-reading provider v3 output and avoid destructive fixes on flaky API responses.
> 
> The QuickNode REST provider is capped at **`rate_limit = 1`** to reduce concurrent refresh failures. A shared **`terraform_state_resource_id`** helper in `common.sh` picks exactly one four-space-indented resource `id` from `terraform state show`, ignoring nested IDs in `api_data`/`api_response`. **`fix-webhook-state.sh`** and the listener **replacement provisioner** both use it; the repair script **fails closed** on parse errors, curl failures, non-404 HTTP codes, and only offers `state rm` after every webhook checks cleanly (with bounded curl retries).
> 
> Adds a **shell regression suite** (`fix-webhook-state.test.sh`) plus a **Vitest wrapper** in the onchain handler, and wires that suite into **agent quality gate** routing and **alerts/AGENTS.md** / runbook docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0d0ff86caeb13114a9d8ea01babd0f7a271a026. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chapati23:
Addressing the two informational P3 notes in the Claude review:

1. Won't fix: the replacement provisioner now intentionally fails closed when Terraform's canonical webhook ID is absent or ambiguous. Continuing would make an inconclusive read look safe and reintroduce the recovery hazard this PR fixes. The runbook documents the blocking behavior, and the regression suite covers it.
2. Won't fix: `get-project-vars.sh` reads the unrelated scalar attributes `email` and `project_id` from different Terraform resource types. Those fields do not share the `restapi_object` nested/top-level `id` collision, so routing them through the webhook-ID parser would conflate separate data shapes without improving safety.

Neither item is deferred work: the first is the intended invariant, and the second is outside the affected parser class with no demonstrated defect.

Issue #1526, opened 2026-07-23
Autofix diff guard: whitespace-filename bypass of the credential-content scan


chapati23:
## The Problem

The autofix diff guard's credential-content scan can be bypassed by a
trailing/leading-whitespace filename, letting an agent-authored file with
credential-shaped content reach the opened PR unscanned. Pre-existing (predates
the #1373 job split; surfaced by the split's adversarial red-team).

Mechanism (verified empirically):
- `diff-trees` reports the real filename, e.g. `leak.ts ` (trailing space).
- `evaluateDiffGuard` / the `guard` CLI **trim** each changed path
  (`scripts/sentry-autofix-finalize.mjs` ~277-279, 609-612), so
  `filesWithCredentialShapedContent` `lstat`s the *trimmed* `work/leak.ts`, hits
  `ENOENT`, and **skips the scan** (~230-236) — the guard returns `{"ok":true}`.
- The workflow's byte-copy loop reads the path **untrimmed**
  (`while IFS= read -r f … done`) and `cp`s the real `work/leak.ts ` into the
  pushed branch.

Impact is now bounded by #1373: the agent job is read-only, so the only
exfiltratable GitHub credential is the read-only `github.token` (still poor
hygiene to publish); the App token is never in the agent job.

## The Solution

Make the guard scan the **exact** path from `changed.txt`, not a trimmed copy —
options:
- Drop the `.trim()` on the path in `evaluateDiffGuard` / the `guard` CLI so the
  credential scan and symlink `lstat` see the same bytes the byte-copy uses; or
- Reject any changed path with edge whitespace or control characters outright
  (a filename with surrounding whitespace is never a legitimate source file), in
  `evaluateDiffGuard`, with a unit test.

Add a `sentry-autofix-finalize.test.mjs` case: a `"leak.ts "` entry containing a
`ghs_`-shaped string must make the guard refuse.

Refs #1373.

chapati23:
Agent claim: claude claimed #1526 for implementation.

Claim ID: claude-20260723T182135-5y5jxk
Branch: main
Claimed at: 2026-07-23T18:21:35.098Z

chapati23:
Moved to review: #1526 is now represented by PR #1556.
Branch: main

Issue #1506, opened 2026-07-22
Autofix marker guard: bind to verdict generation (ABA) in the deterministic finalize job


chapati23:
## The Problem

The autofix finalize step guards the terminal `sentry:fix-pr-opened` marker by
re-reading `sentry:verdict-code-fix` immediately before and after the write
(#1389). That guard tests **label presence**, not **verdict identity**. An ABA
interleaving defeats it:

1. Autofix opens a PR for verdict generation V1.
2. A regression re-queues the issue; ingest sheds `sentry:verdict-code-fix`.
3. A re-triage run re-applies `sentry:verdict-code-fix` (a NEW generation V2,
   based on the regressed occurrence) — all within the finalize window.
4. The post-write re-read sees the label present → marks the stub fixed with
   V1's diff, which never addressed V2's regression.

In practice this is extremely narrow: a full triage-agent run (minutes, twice
daily) cannot realistically complete inside a single autofix run's seconds-to-
low-minutes marker window, and the schedules don't overlap. But the guard is a
proxy, not a proof.

## The Solution

Bind the guard to a **verdict generation token** rather than label presence.
Options:

- Capture the databaseId/createdAt of the verdict comment the diff was based on
  at agent dispatch, thread it into finalize, and require the live newest
  verdict comment to still be that one before writing the marker.
- Reuse the existing regression-fence machinery (`selectVerdictComment` +
  regression-comment timestamp comparison) so "verdict newer than the regression
  it must investigate" is asserted against the SAME comment the diff used.

## Where

This belongs in the **deterministic finalize job** from the #1373 job-split
(#1490), where the verdict snapshot is an explicit, validated input rather than
re-read from mutable labels. Implement generation-binding there instead of
bolting it onto the inline finalize step #1500 replaces.

Refs #1389, #1373.

chapati23:
## Round-5 review (PR #1500) widened the scope

Two more P1s on #1500 both funnel into the **same root**, which is what makes
this issue the real fix rather than more inline guards:

- **Reconciliation reapplies the marker to a stale diff.** `openAutofixPrExists`
  in `sentry-autofix-select.mjs` routes a stub with a fresh verdict but an open
  deterministic-branch PR to the no-agent reconcile path, which reapplies
  `sentry:fix-pr-opened` to the EXISTING PR's diff. If that PR's diff came from a
  now-superseded verdict generation (shed then restored by re-triage), the
[...]

Issue #1389, opened 2026-07-21
Autofix marker write races ingest regression-reopen (stale fix-pr-opened suppresses re-fix)


chapati23:
### Problem

In the Phase 2b autofix leg (PR #1374), the fix job re-reads the queue stub's labels immediately before push/PR-create and refuses if `sentry:verdict-code-fix` is gone (regression re-queue guard). But a TOCTOU window remains **after** that check: if the ingest workflow reopens a regression between the pre-push label read and the outcome write (label application), it sheds the verdict + autofix markers and adds `sentry:needs-triage` — then the autofix step adds `sentry:fix-pr-opened` to the now-needs-triage occurrence. Once re-triage reapplies `sentry:verdict-code-fix`, the selector skips the stub because of that stale terminal marker, so the regressed occurrence never gets another autofix attempt.

The two workflows (ingest, autofix) mutate the same stub's labels on independent concurrency groups with no atomic compare-and-set, so a pure-narrowing re-check cannot fully close it. Codex also flagged auditing the analogous `sentry:fix-refused` marker-write path for the same race.

### Done means

- The `sentry:fix-pr-opened` (and `sentry:fix-refused`) marker is applied only while the occurrence that was validated is still the current one — e.g. re-read `sentry:verdict-code-fix` immediately before the marker write and skip it if absent, and/or make the reconcile path re-validate before re-applying. Document the residual window if it cannot be fully eliminated without cross-workflow locking.
- A test covering the regression-mid-run ordering.

### Notes

Low probability (needs a regression to land in a ~seconds window during PR creation); no data loss (the fix PR still exists and is human-reviewed), but the regressed occurrence's re-fix is silently suppressed until a human clears the marker. Deferred from PR #1374 to keep that PR's review scope bounded.

chapati23:
Agent claim: codex claimed #1389 for implementation.

Claim ID: codex-20260721T174809-b9ve1q
Branch: work
Claimed at: 2026-07-21T17:48:09.417Z

PR #1527, opened 2026-07-23
fix(sentry-autofix): isolate the fix agent in a read-only job, hand off to a trusted finalize job (#1373)

### LINKED ISSUE #1373: Sandbox claude-code-action agents against process-env credential reads (autofix + claude.yml)
## The Problem

`.github/workflows/sentry-autofix.yml` (ADR 0036 Phase 2b) runs a `claude-code-action` agent over an UNTRUSTED, second-order Sentry triage verdict, then automatically publishes the agent's edits as a PR. The action holds its `github_token` input and the `CLAUDE_CODE_OAUTH_TOKEN` in the Claude CLI process env, and the agent's `Read` tool is unrestricted. A prompt-injected verdict could therefore attempt to read a process-environment surface (e.g. `/proc/self/environ`) and copy a credential into an edited file or the summary, which the deterministic finalize step publishes; the diff guard only checks path/count, and required CI + human review happen after the text is already public.

This exposure is inherent to `claude-code-action` and identical to the repo's existing `.github/workflows/claude.yml` agents. The autofix workflow already removes the ambient `GH_TOKEN` from the agent's env (verdict is pre-fetched, no `gh` tool) and denies all code execution, but the action's own token handling remains. The OAuth token is inference-only and rotatable, and the pipeline is inert until an operator provisions the App and flips `SENTRY_AUTOFIX_ENABLED`.

## Solution ideas

- Run the autofix `claude-code-action` step in an OS/filesystem sandbox that denies process/environment reads (seccomp/AppArmor profile, or a container job with a restricted `/proc`), so the agent cannot reach credential surfaces.
- Or mint a dedicated, minimally-scoped, short-lived Claude credential for the autofix leg so a leak doesn't compromise the shared `claude.yml` token.
- Track whether `claude-code-action` gains a first-class way to withhold its auth token from tool subprocesses.

Applies equally to `claude.yml`; consider a repo-wide agent-sandbox baseline. Surfaced by an independent (Codex) review of the Phase 2b autofix PR. Refs #1278.

chapati23:
## The Problem

- The Sentry autofix leg is merged but inert; the last activation gate is #1373 — **agent credential isolation**.
- Today the untrusted LLM fix agent runs in the same `autofix` job that holds `issues: write` and mints the GitHub **App token**. Its `claude-code-action` process env carries a **write-capable** `github.token`, reachable in principle via `/proc/self/environ` — so a prompt-injected agent could write issues.

## The Solution

Split the single job into two, so the agent never runs in a job with write credentials:

- **`agent`** (`contents: read` + `issues: read` only, `max-parallel: 2`) — runs `claude-code-action`, then a live-FS symlink tripwire, then hands its **whole working tree** to finalize as an artifact.
- **`finalize`** (`issues: write` + App token, `max-parallel: 1`, `needs: [select, agent]`) — downloads the tree, **re-derives** the diff + guard against its **own** pristine clone (trusting no agent-provided metadata), then does every write: refused analysis, PR open, the #1389 marker guards, reconcile.

Now an agent that exfiltrates its job's `github.token` gets a **read-only** token — it cannot write issues, push, or open PRs. The verdict logic is unchanged.

## Details

- Handoff uploads the **full** tree (not a diff/list) with `include-hidden-files: true` + `if-no-files-found: error`, per-entry name `sentry-autofix-worktree-<issue>`, 1-day retention; finalize re-derives so it trusts only raw file bytes that pass its own guard. `upload-artifact` v7.0.1 / `download-artifact` v7.0.0, SHA-pinned (auto-tracked by the existing github-actions Dependabot group).
- `finalize`'s `if: !cancelled()` + `fail-fast: false` so one entry's failed agent doesn't skip the whole finalize matrix; a missing artifact fails only that entry.
- The Claude OAuth **inference** token remains in the action env (inherent to `claude-code-action`; worst case is inference-quota abuse) — explicitly **out of scope** for #1373.
- The agent's `Read,Grep,Glob,Edit,Write,MultiEdit` tool set (no shell) cannot create symlinks, so the symlink layers are defense-in-depth against an unreachable primitive; confirmed by an adversarial red-team.

## Validation

- `actionlint` clean; `bash -n` on every run block; `check-autofix-ci-trust.mjs` (29 workflows) clean; `docs:navigation-eval` 33/0; finalize + select unit tests pass.
- **Adversarial red-team** (5 opus lenses → refute each finding): **0 defects introduced** by the split. Two pre-existing items surfaced and triaged: the refused-path stale-verdict asymmetry (justified — publishes only a diagnostic comment, not a mergeable PR) and a whitespace-filename credential-scan bypass (pre-existing, harm reduced by this split, tracked in #1526).
- Not enabled: `SENTRY_AUTOFIX_ENABLED` stays off; end-to-end runs only after the operator flip.

## Deferrals

- Verdict-generation ABA check (finalize refusing a shed-then-re-added verdict) — #1506, the stacked follow-up that builds on this job's deterministic finalize.
- Pre-existing whitespace-filename bypass of the guard's credential-content scan — #1526.

Closes #1373. Supersedes #1490.

### DIFF
--- a/.github/workflows/sentry-autofix.yml
+++ b/.github/workflows/sentry-autofix.yml
@@ -2,10 +2,11 @@ name: Sentry Autofix
 
 # Stage C / Phase 2b of the Sentry triage pipeline (ADR 0036): for queue stubs
 # verdicted `code-fix` whose parsed `affected_repo` is EXACTLY this repo, a
-# claude-code-action agent implements a scoped fix in a fresh checkout, then a
-# DETERMINISTIC finalize step commits, pushes a branch, and opens a PR via a
-# GitHub App token. PR-ONLY — nothing ever merges automatically; merge stays
-# human (ADR 0036).
+# claude-code-action agent implements a scoped fix in an isolated READ-ONLY
+# `agent` job (issue #1373), which hands its whole working tree to a separate
+# TRUSTED `finalize` job that re-validates the diff, commits, pushes a branch,
+# and opens a PR via a GitHub App token. PR-ONLY — nothing ever merges
+# automatically; merge stays human (ADR 0036).
 #
 # Prompt-injection containment is the whole design point (the verdict is
 # second-order UNTRUSTED Sentry data):
@@ -22,8 +23,10 @@ name: Sentry Autofix
 #     diff guard + required CI + human review catch. (Residual: claude-code-action
 #     itself holds its `github_token` input + the Claude OAuth token in the CLI
 #     process env, which the agent's Read tool could in principle reach via a
-#     process-env surface — inherent to the action, same as claude.yml, tracked
-#     as a sandboxing follow-up.)
+#     process-env surface. Because the `agent` job is READ-ONLY (contents:read +
+#     issues:read, issue #1373), an exfiltrated `github_token` cannot write
+#     issues, push, or open PRs; only the Claude OAuth inference token remains —
+#     inherent to the action, tracked as a sandboxing follow-up.)
 #   - NO git AT ALL against the agent-tainted checkout, and NO write credential
 #     in the LLM step. The agent has Write access to the checkout INCLUDING
 #     .git/, so any git command there would run agent-planted hooks / clean-smudge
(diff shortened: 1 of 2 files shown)


chapati23:
Thanks for the thorough trace. Confirming the one P3 (finalize depending on the whole `agent` job rather than per-entry) is a known, accepted inherent GitHub Actions job-graph constraint — it's called out as a risk in the design and handled by `finalize`'s `if: !cancelled()` + `fail-fast: false` so a slow/failed entry never strands a sibling. No change needed, as noted.

Issue #1444, opened 2026-07-22
Implement peg monitoring for oracle-less stablecoins (ADRs 0042–0045, phases 2–4)


chapati23:
### Goal

Implement the peg-monitoring system specified in `docs/PLAN-peg-monitoring.md` and ADRs 0042–0045, so a real EUROP/EUR depeg produces a trustworthy page with a decision package before the EURm/EUROP FPMM drains reserve at par.

### Context and links

- Design + evidence: `docs/PLAN-peg-monitoring.md` (live-verified 2026-07-22 market structure; two adversarial review rounds)
- Decisions: ADR 0042 (bridge-hosted external price poller), 0043 (service-local registry), 0044 (gated thresholds plane), 0045 (executable-sell-price paging semantics)
- Existing surface: `docs/notes/polygon-monitoring.md` (EURm/EUROP pool, MANUAL feed, migration-multisig breaker path)

### Acceptance criteria

- [ ] **Phase 2 — bridge peg module:** registry schema + `peg-registry.json` + fixtures; Kraken/Bitvavo adapters with the `{vwap, filledFraction, capped}` observation contract; structural `SwapEvent` companion queries (bounded page + saturation flag, no `_aggregate`); `mento_peg_*` gauge family in its own module/loop (CDP-precedent lifecycle, never gates `/health`); registry referential-integrity script wired into the quality gate and CI.
- [ ] **Phase 3 — alerts stack:** `alerts/rules/peg-thresholds.json` + `jsondecode`/`for_each` rule group with freshness gating on every rule, deliberate `no_data_state`, duration-fraction sustain; routing warn→Slack / critical→page; gated apply only after producer telemetry is live (follow the no-data rollout discipline from `docs/notes/polygon-monitoring.md`).
- [ ] **Phase 4 — operations:** dashboard decision-package panel; onboarding runbook (census, coverage-class gate, signer SLA, trading-limit verification); scheduled re-census with registry-rot and critical-path-unreachable alerts.

### Design-review carryover (validated PR #1445 findings, implementation-owned)

These review findings from the ADR PR are correct but implementation-level; the ADR decisions stand and these land as code/test contracts, not further ADR text:

[...]

PR #1603, opened 2026-07-24
feat(alerts): add producer-backed peg listing alerts

chapati23:
## The Problem

- The Peg producer can confirm when a required market stays missing, but Grafana does not yet turn that evidence into an operator alert.
- Registry, critical-path, and indexed-pool failures need explicit alerts with safe routing and rollout instructions.
- Policy activation must fail closed without offering an unsafe shortcut for later policy rollovers.

## The Solution

- Add warning-only Grafana alerts that consume the producer's exact policy version, bounded absence streak, and fresh authoritative timestamps.
- Keep all nine Peg alert resources absent by default behind one shared `peg_alerts_enabled = false` guard.
- Document a protected first activation and block later asset or source rollovers until a separate publication or per-policy consumer activation boundary exists.

## Details

- Build on the producer merged in #1601; this PR's current-main delta is 17 files.
- Alert when a required market is absent for the policy's configured number of successful checks; never infer streaks from scrape history.
- Add Registry Rot checks for non-deep sources, Critical Path Unreachable checks for required deep-book sources, and Indexed Pool Unreachable checks for configured DEX pools.
- Keep these operational alerts at `for = "0s"` with `no_data_state = "OK"`, and route them directly to `#alerts-infra` without PagerDuty or Splunk escalation.
- Add active and retained-previous policy rule sets with exact version selectors and an exact legacy-policy compatibility default.
- Register the nine new Grafana resources by exact file and address in the production-infrastructure identity contract; unknown resources remain rejected.
- Require policy publication and producer evidence while the global guard is false, then a reviewed source flip and protected full apply for the first activation.
- Forbid targeted apply and guard disablement as rollover tools. Disabling the shared guard after activation would remove all live Peg consumers.
- Treat each due configured-source poll as exact-pair revalidation. It does not discover or mutate source topology.

## Invariants

- Only a successful authoritative exact-pair `absent` result advances the producer-owned absence streak; `listed` and `halted` reset it, while unknown or failed checks preserve prior evidence.
- Active and retained-previous rules select one exact policy version and never mix versions.
- Missing or stale listing evidence is not delisting.
- Listing and reachability alerts remain warning-only; the heartbeat rule owns total producer-loop outage.
- New Terraform resource addresses must be explicitly registered. The checker still fails closed on every unregistered resource or data block.

## Degraded Behavior

- Listing and reachability rules use `no_data_state = "OK"` so missing producer evidence does not become a false delisting alert.
- Producer heartbeat and blindness rules retain their existing fail-closed behavior for real loop or pricing outages.
- Retained-previous rules stay active until a later reviewed policy cleanup removes them.

## Intentional Non-goals

- This PR does not apply Terraform, deploy the producer, activate the protected policy artifact, or prove live firing and recovery.
- It does not change paging thresholds or route listing and reachability warnings to Splunk On-Call.
- It does not add an independent scheduled re-census executor.
- It does not enable post-activation policy rollovers; #1444 remains blocked on a safe publication or per-policy activation boundary for those transitions.

## Validation

- `CI=true pnpm agent:quality-gate --run` — all mapped alert, PromQL, registry, identity-contract, documentation, Trunk, and Terraform checks passed at `c187832c`.
- The focused Terraform guard test passed with all nine resources absent at the default and present when enabled.
- The guard mutation test, alert-rule lint tests, PromQL metric cross-check, and registry integrity tests passed.
- Fresh-context semantic autoreview found no actionable issue at `c187832c` with 0.99 confidence; deterministic local autoreview was also clean.
- No deployment, Terraform plan, Terraform apply, or alert-delivery test ran.

## Deferrals

- #1444 — protected first activation, live firing and recovery proof, an independent scheduled re-census executor, and a safe boundary for later policy rollovers remain.

Refs #1444

## Checklist

- [x] The Solution is understandable before the implementation details.
- [x] Every knowing deferral links a GitHub issue.
- [x] Architecture decision? No new decision; this implements and updates ADRs 0043, 0044, and 0049 while preserving ADRs 0047 and 0048.

## Ship Checklist

- [x] ship skill used
- [x] local review and gate run
- [x] PR readiness probe planned


chapati23:
Heads-up: current main is red because the Peg Terraform resources from #1581 are present but not registered in the production-infrastructure identity contract. #1618 restores main by registering those exact merged resources plus the dashboard metrics_bridge_url sink from #1588. This overlaps your surfaces.mjs additions; please refresh from main after #1618 lands and drop the duplicate registry delta.

chapati23:
@codex review

chapati23:
@codex review

Issue #1586, opened 2026-07-24
Sentry-triage URL validators don't reject Slack link-control chars (<, >, |)


chapati23:
## The Problem

- The Sentry-triage pipeline splices validated URLs directly into Slack mrkdwn link syntax (`<url|text>`) in several places — e.g. `link()` and `idAndProject()` in `scripts/sentry-triage-digest.mjs`.
- The shared validators that gate those URLs (`isSafeSentryPermalink` / `extractPermalink` in `scripts/sentry-triage-project-core.mjs`, and the digest's own `isHttpsUrl`) only check protocol (`https:`) and hostname (`*.sentry.io` or `github.com`) — they never reject or escape Slack's link-control characters (`<`, `>`, `|`) if one somehow appeared inside the URL itself.
- Today this can't actually be triggered: Sentry doesn't emit permalinks containing those characters, and GitHub issue/PR URLs are structurally clean. But the validators don't *structurally guarantee* that, so a future change to either upstream format (or a different untrusted URL source added to the pipeline later) could silently break a Slack block or splice unintended link syntax.
- Surfaced independently by both the local `/review` pass and the Codex review during PR for the digest Slack-format rework (branch `fix/sentry-triage-digest-slack-format`); both agreed it's a real gap but out of scope for that formatting-only change.

## The Solution

Harden the shared URL validators (`isSafeSentryPermalink` in `scripts/sentry-triage-project-core.mjs`, and the digest's `isHttpsUrl`/`link()` in `scripts/sentry-triage-digest.mjs`) to also reject any URL containing `<`, `>`, `|`, or control/whitespace characters before it's treated as a safe link target — fail closed (treat as unsafe/omit the link) rather than splice it in. Grep for every `<${url}|` / `` `<${...}|${...}>` `` splice site across the sentry-triage scripts to make sure all of them route through the hardened validator, not just the ones touched here.

chapati23:
Agent claim: claude claimed #1586 for implementation.

Claim ID: claude-20260725T080333-o99j94
Branch: main
Claimed at: 2026-07-25T08:03:33.980Z

chapati23:
Moved to review: #1586 is now represented by PR #1615.
Branch: main

PR #1615, opened 2026-07-25
fix(sentry): reject Slack link-control + control chars in triage URL validators (#1586)

chapati23:
## The Problem

- The Sentry-triage URL validators accept `<`, `>`, `|`, ASCII control chars, and
  whitespace in a URL's path/query — all structurally valid via `new URL()` — and
  return the **raw** input string.
- Those values flow into the digest's Slack mrkdwn link splices (`<url|text>`),
  where `<`, `>`, `|` are Slack's link-control characters. A permalink carrying
  them can break out of the link and spoof the display text or split the block —
  a link-injection surface (issue #1586).

## The Solution

Harden every URL-shape validator across the class with a single shared guard on
the raw input, failing closed (the URL is dropped/omitted, never spliced):

```js
const UNSAFE_URL_CHARS = /[<>|\x00-\x20\x7f]/;
```

Rejected: `<` `>` `|` (Slack link-control), ASCII control `\x00–\x1f` (incl. tab,
newline, CR), space, and DEL `\x7f`. Legit Sentry/GitHub URLs percent-encode all
of these, so nothing valid is rejected.

## Details

- **All sites covered** (the sweep found one Slack sink — `sentry-triage-digest.mjs`
  — and every validator that feeds it):
  - `isSafeSentryPermalink` in `sentry-triage-project-core.mjs`, `sentry-triage-ingest.mjs`, `sentry-triage-archive.mjs`
  - `isHttpsUrl` (the digest sink choke point for every `link()` / `idAndProject()` splice) **and** `isGithubUrl` (projected/autofix arrow links) in `sentry-triage-digest.mjs`
- **Intentionally not changed:** `isSafeNextPageUrl` (ingest) guards an
  authenticated Sentry pagination fetch (SSRF/token concern), never a Slack
  message — out of scope.
- Tests: direct reject/accept per validator, plus an end-to-end `buildDigest`
  test that poisons a queue URL and asserts it is dropped (no `<…|` breakout),
  with a positive control that clean URLs still render as links. Control-char
  fixtures use `\x00`/`\x7f`/`\n` escapes so no literal control byte lands in
  source.

## Validation

- Sentry test suites: project **91**, digest **62**, ingest **56**, archive **37** — all pass (incl. the new `(#1586)` cases).
- `eslint` + `trunk fmt` clean on all 7 files.
- Adversarial recheck: `<`,`>`,`|`,tab,newline,space rejected; percent-encoded (`%3C`) and clean permalinks pass.

## Deferrals

- None.


chapati23:
@codex review

PR #1643, opened 2026-07-27
[PROBE - DO NOT MERGE] non-admin env-secret gating test

gisk0:
this is a throwaway security probe that will be closed and its branch deleted after one workflow run.


chapati23:
Closing — the probe workflow's YAML indentation was flattened in transcription (invalid → the run failed on a parse error, not the environment gate), and the account used has admin, so this didn't test the non-admin case. Re-running a corrected version. Branch will be deleted.

PR #1707, opened 2026-07-29
test(scripts): score budget cases against an injected fixture corpus

### LINKED ISSUE #1701: Decouple the navigation scoring tests from live documentation sizes
## The Problem

`an over-budget answer is measured and fails` in `scripts/docs-navigation-eval.test.mjs` is a unit test of the **scoring logic**, but it can only construct its over-budget case out of **real documentation files**. That coupling has now produced three separate rounds of work, none of which removed it:

1. #1690/#1691 slimmed the agent entry points. `agent-quality-gate-mechanics.md` lost 12,993 bytes against a 7,181-byte margin, the fixture stopped clearing the budget, `questions_over_budget` went to 0, and the test began asserting the opposite of its name. `main` went red and every PR failed the required `Lint + test root scripts` check.
2. The repair added a third hardcoded document to restore the margin — which re-arms the same trap for the next person who trims a file, and additionally couples the test to `docs/notes/sentry-triage-pipeline.md`, a file an unrelated in-flight PR grows.
3. #1700 replaced the hardcoded list with runtime derivation. Better, but the review then found two drift bugs *within the derivation itself*: it sorted by worktree bytes while the scorer reads bytes at `repository_base_commit`, and cheapest-per-file selection quietly consumed suite-wide unique-byte headroom that later documentation PRs need.

Every one of these is a way of *managing* the coupling. The failure keeps recurring because a scoring-logic test should not depend on how large the repository's prose happens to be this week.

## Solution ideas

The blocker is `validateLoadedSources` (`scripts/docs-navigation-eval-result.mjs:265-285`): it re-reads each declared source at `repository_base_commit`, rejects a mismatched `bytes`/`sha256`, and accumulates the **real** `content.length`. That is correct and load-bearing for the real eval — a result must not be able to lie about what it read — so a fabricated source cannot be used by the test today.

Options, roughly in order of preference:

- **Inject the source reader.** Let the scorer take a source-resolver dependency that defaults to the real git read. Tests pass a fixture resolver with synthetic paths and sizes; production behaviour is unchanged and still verifies against the repository. This decouples every size-dependent test at once, not just this one.
- **Fixture corpus.** Point the test at a small committed fixture tree under `docs/evals/` whose sizes are stable by construction and owned by the eval, rather than at live documentation.
- **Assert on the scorer, not on the corpus.** Split the concern: unit-test the budget comparison against synthetic numbers, and keep a separate, coarser integration check that the real corpus is within budget. The first is what the failing test actually wants to prove.

Whichever is chosen, the acceptance bar is the same: trimming or growing any documentation file must not be able to change the meaning of a scoring-logic assertion.

## Done means

No test of scoring logic reads the size of a live documentation file. Verified by trimming a large canonical doc locally and confirming that no assertion in `docs-navigation-eval.test.mjs` changes outcome.

Refs #1700

chapati23:
## The Problem

- `an over-budget answer is measured and fails` is a unit test of **scoring logic**, but it could only build its case out of **real documentation files** — so editing prose could silently change what it asserts.
- That coupling took `main` red once (#1690/#1691 trimmed a doc past the fixture's margin) and produced three rounds of repair, two of which introduced their own drift bugs inside the fix.
- Every repair so far *managed* the coupling. None removed it.

## The Solution

The scorer now takes its source reader as a dependency, defaulting to the real git read. Production keeps re-reading the repository and keeps rejecting any result that misreports what it loaded; the test passes a synthetic corpus instead, so no documentation file's size can reach a scoring assertion.

The fixture suite still supplies routes, ids, and targets — only the bytes are chosen. That buys an assertion live sizes could never express: an answer at **exactly** 45,000 bytes is not over budget, and 45,001 is.

## Details

- `scoreNavigationResult` accepts `readSource`, defaulting to `readSourceAtCommit`. It threads into `historicalInventoryMap` and both `validateLoadedSources` call sites — the only three places the scorer touches bytes. **22 lines of production diff.**
- **The production guarantee is unchanged.** `docs-navigation-eval.mjs --validate`, `assertPassingNavigationResult`, and `validateHistoricalBaseline` never pass a resolver, so `validateLoadedSources` still reads each declared source at `repository_base_commit`, still rejects a mismatched `bytes` or `sha256`, and still accumulates the real `content.length`. A result still cannot lie about what it read.
- The issue's fallback (a committed fixture corpus under `docs/evals/`) was not needed — injection preserved the guarantee, so the cheaper option was not taken at the cost of the stronger one.
- Guarantees kept: the clean-baseline assertion, `assert.deepEqual(scored.errors, [])` on all three cases, the `total_unique_source_bytes` isolation check, and an explicit not-constructible failure that now names the answer, the bytes already spent, and the budget.

## Validation

**Drift demonstration** — the deliverable, not just a green suite. In an isolated clone, `docs/deployment.md` was trimmed 35,484 → 261 bytes, committed, and `origin/main` repointed at it so the scorer genuinely read the trimmed file. That is 2.7× the trim that took `main` red.

| | before trim | after trim |
| --- | --- | --- |
| `docs/deployment.md` at base commit | 35,484 | **261** |
| live-corpus `total_unique_source_bytes` | 208,178 | **189,116** |
| fixture at-budget `source_bytes` / over-budget count | 45,000 / 0 | 45,000 / 0 |
| fixture over-budget `source_bytes` / count | 45,001 / 1 | 45,001 / 1 |
| fixture `total_unique_source_bytes` / errors / passed | 65,481 / 0 / false | 65,481 / 0 / false |

The live corpus moved 19,062 bytes; every scoring number is byte-identical.

Run against the **pre-change** test for contrast, the same trim changed both the document set it selected and the value it measured — the coupling this removes.

- `node --test scripts/docs-navigation-eval.test.mjs` → 34 pass, 0 fail
- `pnpm docs:navigation-eval -- --check-fixtures` → `valid:true`, `total_unique_reserve_surplus_bytes: 21054`
- `pnpm agent:quality-gate --run` → all mapped commands passed
- **Mutation-checked**: disabling the over-budget comparison fails `0 !== 1`; changing `>` to `>=` at the boundary fails `1 !== 0`, which is the off-by-one only a chosen-size corpus can catch.

## Deferrals

- None

Closes #1701

---

> [!NOTE]
> **Low Risk**
> Production scoring behavior is unchanged when `readSource` is omitted; only test harness and an optional dependency for byte resolution were added.
> 
> **Overview**
> **Navigation eval scoring** can now resolve declared source bytes through an optional `readSource` hook on `scoreNavigationResult`, threaded into `historicalInventoryMap` and `validateLoadedSources`. Production callers keep the default `readSourceAtCommit`, so byte/`sha256` checks against `repository_base_commit` are unchanged.
> 
> **Budget tests** no longer build over-budget cases by loading extra real documentation (`loadSourcesUntilOverBudget` removed). They inject a synthetic corpus via `readSource`, with helpers to fabricate canonical markdown at chosen sizes and to score the first answer at exactly the suite limit, at the limit, or one byte over—so trimming or growing a live doc cannot flip the test meaning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6269da759f8578745059454a2ab740e21dd4cf51. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/docs-navigation-eval-result.mjs
+++ b/scripts/docs-navigation-eval-result.mjs
@@ -148,13 +148,19 @@ function fixtureSourcePaths(suite, questions = suite?.questions ?? []) {
   return paths;
 }
 
-function historicalInventoryMap(repoRoot, commit, paths, errors) {
+function historicalInventoryMap(
+  repoRoot,
+  commit,
+  paths,
+  errors,
+  readSource = readSourceAtCommit,
+) {
   const records = new Map();
   for (const file of paths) {
     if (!isDocumentationPath(file)) continue;
     let content;
     try {
-      content = readSourceAtCommit(repoRoot, commit, file);
+      content = readSource(repoRoot, commit, file);
     } catch {
       continue;
     }
@@ -224,6 +230,7 @@ function validateLoadedSources({
   errors,
   repoRoot,
   baseCommit,
+  readSource = readSourceAtCommit,
 }) {
   if (!Array.isArray(sources) || sources.length === 0) {
     errors.push(`${label} must list at least one loaded source`);
@@ -264,7 +271,7 @@ function validateLoadedSources({
     }
     let content;
     try {
-      content = readSourceAtCommit(repoRoot, baseCommit, source.path);
+      content = readSource(repoRoot, baseCommit, source.path);
     } catch (error) {
       errors.push(error instanceof Error ? error.message : String(error));
       continue;
@@ -563,11 +570,17 @@ export function validateNavigationResultShape(
   return errors;
 }
 
+// `readSource` resolves one declared source to its bytes at the evaluated
+// commit. The default reads the repository, which is what makes a result unable
+// to lie about what it loaded, so production callers must leave it alone. Tests
+// of the scoring arithmetic pass a fixture corpus instead, so their sizes are
+// chosen rather than inherited from whatever the documentation weighs today.
 export function scoreNavigationResult({
   suite,
   result,
   repoRoot,
   questionId = null,
+  readSource = readSourceAtCommit,
 }) {
   const evaluatedQuestions = questionId
     ? suite.questions.filter((question) => question.id === questionId)
@@ -635,6 +648,7 @@ export function scoreNavigationResult({
     sourceCommit,
     resultSourcePaths(suite, result, evaluatedQuestions),
     errors,
+    readSource,
   );
   if (run.fresh_context !== true)
     errors.push("result.run.fresh_context must be true");
@@ -647,6 +661,7 @@ export function scoreNavigationResult({
     errors,
     repoRoot,
     baseCommit: sourceCommit,
+    readSource,
   });
   const expectedBootstrap = new Set(suite.bootstrap_sources);
   if (
@@ -706,6 +721,7 @@ export function scoreNavigationResult({
       errors,
       repoRoot,
       baseCommit: sourceCommit,
+      readSource,
     });
     for (const file of loaded.paths) {
       if (bootstrap.paths.has(file)) {
--- a/scripts/docs-navigation-eval.test.mjs
+++ b/scripts/docs-navigation-eval.test.mjs
@@ -159,86 +159,177 @@ function score(result) {
   });
 }
 
-// The scorer sums an answer's context spend from the real bytes each loaded
-// source has at repository_base_commit, and rejects reported bytes or sha256
-// that disagree with that file, so a synthetic oversized source cannot push an
-// answer over budget — only real documentation can. Two rules keep that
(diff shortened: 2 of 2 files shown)


chapati23:
Review acknowledged — LGTM with no findings, nothing to address.

For the record, the load-bearing check on this one is the drift demonstration rather than the green suite: `docs/deployment.md` was trimmed 35,484 → 261 bytes in an isolated clone with `origin/main` repointed at it, so the scorer genuinely read the trimmed file. The live corpus moved 19,062 bytes and every scoring number stayed byte-identical. The same trim against the pre-change test changed both the document set it selected and the value it measured.

Issue #1706, opened 2026-07-29
Sentry triage: a lost close response can strand a closed stub carrying sentry:needs-triage


chapati23:
## The Problem

Two sites in `.github/workflows/sentry-triage-agent.yml` restore `sentry:needs-triage` after a `gh issue close` **reported** failure:

- `:357` — the matrix close-failure path
- `:506` — the project close-failure path

Normally that is correct: the close failed, the stub is still open, and re-labelling makes it selectable again. But if the close actually landed server-side and only its **response** was lost, the label goes onto a stub that is now closed — producing the pairing nothing in the pipeline can see:

- the triage selector is `gh issue list --label sentry-triage --label 'sentry:needs-triage' --state open`, so a closed stub is invisible to it;
- ingest skips any stub whose state is not open, and its regression path only reopens when `lastSeen` postdates `closedAt`.

A queue issue can therefore sit closed indefinitely wearing `sentry:needs-triage`, seen by no stage.

This is the same class #1696 spent its later review rounds on — **a rejected command is not proof its remote mutation did not happen** — surfacing here in the triage workflow rather than the archive scripts. #1696 fixed the archive-side instance (the live-regression refusal now re-reads and reopens from observed state); these two were found by the audit that fix prompted, and were deliberately left out of that branch because they are a different file with a different trigger.

## Solution ideas

- After a close reports failure, **re-read the issue** and branch on its observed state rather than assuming the close did not happen: if it is closed, either leave it closed without `needs-triage`, or reopen it so the label is meaningful. Mirrors `reconcileToTarget` in `scripts/sentry-triage-archive.mjs`.
- Or make the pairing self-correcting: have ingest treat "closed + `sentry:needs-triage`" as a recoverable state and reopen it. That fixes every current and future producer of the pairing rather than the two known ones — likely the better fix, and it also covers hand-editing mistakes.
- Whichever is chosen, add a test that models a close whose response is lost, in the shape used by `scripts/sentry-triage-archive.test.mjs` (a stateful mock with an `ambiguousOn` hook that applies the mutation and then throws).

## Done means

No pipeline stage can produce, or leave in place, a closed queue stub carrying `sentry:needs-triage`. Proven by a test simulating a lost close response at each site.

Refs #1371, #1696, #1282

chapati23:
Agent claim: claude claimed #1706 for implementation.

Claim ID: claude-20260729T232912-kjuzto
Branch: main
[...]

Issue #1698, opened 2026-07-29
Split the oversized Sentry triage archive/ingest modules (600-line soft cap)


chapati23:
## The Problem

The Sentry triage scripts are far past the repo's 600-line soft cap (`docs/pr-checklists/recurring-review-patterns.md:102`):

| file | lines (2026-07-30) | when filed |
| --- | --- | --- |
| `scripts/sentry-triage-archive.mjs` | **2125** | ~1366 |
| `scripts/sentry-triage-ingest.mjs` | **1225** | ~1259 |
| `scripts/sentry-triage-project-core.mjs` | **957** | — |

Archive grew ~760 lines after #1696 landed. Note `project-core.mjs` — the proposed destination for shared contracts — is itself at 957 and would become the next instance of this issue. It needs a target of its own, not more arrivals.

Both were already over before #1696; that PR added the CAS, compensation, and freshness-baseline logic on top. The archive module now holds several separable concerns in one file — Sentry API access, the CAS and its compensation, freshness-baseline capture and parsing, audit-comment construction, and the queue-stub settlement state machine. The race and rollback invariants are the hardest thing in this package to audit, and a 1300-line module makes that harder than it needs to be.

## Split along the re-queue seam, not just by concern (added 2026-07-30)

Line count is the symptom. The evidence for the real seam came from #1716, which took **three separate P1s in this area** — none a re-raise, each a distinct interaction between behaviours that were correct in isolation:

1. The archive's live-regression refusal re-queued a stub without the parser's staleness fence, so the previous round's verdict stayed newest-admissible.
2. `reopenQueueIssue` mutated labels *before* posting that fence, so an interrupted run left a queued-but-unfenced stub.
3. After both were fixed: a **failed** regression reopen never entered `reopenedNumbers`, so the same run's stranded-stub sweep recovered it through the deliberately fence-free *bookkeeping* path — laundering a Sentry-evidence re-queue into a bookkeeping one.

The common cause is that "re-queue this stub" is an ad-hoc sequence of label writes and comment posts open-coded at many call sites, and each site independently decides whether it fences. A sweep found 14 such sites. Correctness depends on a rule — *fence when the re-queue is caused by new Sentry evidence, do not when it is bookkeeping* — that lives only in prose and in each author's head. Over-fencing is as wrong as under-fencing: it discards verdicts the pipeline means to keep.

[...]

Issue #1282, opened 2026-07-15
tracker: Sentry issue triage & autofix automation (ADR 0036) — phased rollout


chapati23:
Tracking issue for the Sentry issue triage & autofix automation decided in ADR 0036 (`docs/adr/0036-sentry-triage-pipeline.md`). This issue is a TRACKER — do not claim it directly; claim child issues via `pnpm issue:claim --issue <n>`.

## Decision summary

A staged, triage-first pipeline: a deterministic ingest turns new/regressed Sentry issues (org `mento-labs`, 6 projects → 4 repos) into GitHub queue issues here; read-only Claude agents post structured verdicts; mutations (Sentry archive, fix PRs) are earned phase by phase with measured accuracy. Codex stays the independent PR reviewer. Nothing ever auto-merges. Rationale, alternatives (cloud routines, OpenClaw, Codex-as-engine, Seer), and verified platform constraints are in the ADR.

## Live state (verified 2026-08-10)

| Stage | Flag | Secret | State |
| --- | --- | --- | --- |
| Ingest | `SENTRY_TRIAGE_ENABLED=true` | `SENTRY_TRIAGE_TOKEN` ✅ | **live**, 2×/day every day (cron `30 5`, `30 13`) |
| Triage agent | `SENTRY_TRIAGE_ENABLED=true` | `SENTRY_TRIAGE_TOKEN` ✅ | **live**, weekdays only, once at 07:55 UTC (cron `55 7 * * 1-5`). A stub filed on a weekend has no verdict run scheduled before Monday — a missing weekend verdict is not a defect. |
| Projection | — | `SENTRY_PROJECTION_TOKEN` ✅ | **live** |
| Autofix | `SENTRY_AUTOFIX_ENABLED=true` | `AUTOFIX_APP_PRIVATE_KEY` ✅ | **live**, weekdays — **zero PRs, and one eligible candidate in its lifetime**. The five `sentry:fix-refused` stubs (#1304, #1313, #1316, #1326, #1328) are one Sentry family, not five candidates, and every refusal was correct. Every `code-fix` verdict since 2026-07-16 is `sentry:projected` to a repo this leg has no checkout for, so the no-op runs are an empty queue reporting success. This was a **supply** problem. #1614 shipped its prompt fixes; #1279 remains the cross-repo supply path. #1758 settled the refusal analysis. |
[...]

Issue #1721, opened 2026-07-30
CI never runs the Sentry triage/autofix test suites — the pipeline's ~400 assertions gate nothing server-side


chapati23:
## The Problem

- **CI never runs the Sentry triage/autofix test suites.** Verified against `origin/main`: the `Lint + test root scripts` job runs 28 test invocations and **none** is a `sentry:*:test`. There is no `pnpm sentry:` invocation anywhere in `.github/workflows/ci.yml`.
- **The quality gate is tested but never run.** `ci.yml:974` runs `pnpm agent:quality-gate:test` — which exercises the gate *script*. `pnpm agent:quality-gate --run` appears in no workflow.
- So roughly 400 assertions across `sentry-triage-{archive,ingest,project,digest}`, `sentry-autofix-{select,finalize}`, `sentry-triage-agent-comment` and the new broker are enforced **only** by the local pre-push hook. Server-side, nothing gates them.

## Why this matters more than it looks

Every one of the **seven** distinct defects fixed on #1716 was caught by tests CI does not run — including a fence that let a run go green over a live regression, and an author-fence gap an outsider could trigger on this public repo. The suite that found them gates nothing.

Concretely:

- **Machine-authored `sentry-autofix/*` PRs are the acute case.** That leg opens PRs by itself, and those PRs are precisely the ones no human ran a pre-push hook for. `ci.yml` already treats them specially in a dozen places; they would land with the pipeline's own suite unexecuted.
- A contributor without this repo's hooks configured bypasses the suite entirely.
- It hid a real defect this week. Codex flagged that the broker's credential-boundary test uses `ps -E`, which Linux procps rejects — so the test cannot pass on the runner. CI could neither confirm nor refute it, because **the test never executed there**. A green CI on that PR meant nothing about the property it claims to prove.

## Solution ideas

- **Simplest and most legible: add explicit steps to the `Lint + test root scripts` job**, one per suite, matching the 28 that already live there. Explicit invocations keep the required-check surface readable and fail with a nameable step.
- **More general: run `pnpm agent:quality-gate --run` in CI** and let its changed-path mapping decide. Heavier, and it makes one check stand for many, but it removes the "someone added a suite and forgot the CI step" failure mode permanently — which is exactly how this gap arose.
- Whichever is chosen, **add a meta-test asserting every `sentry:*:test` script in `package.json` is reachable from CI**, so the next suite cannot be added without wiring. Without that, this recurs.

## Done means

[...]

PR #1722, opened 2026-07-30
refactor(sentry): give re-queueing one owner that decides the fence

chapati23:
## The Problem

- "Re-queue this stub" was an open-coded sequence of label writes and comment posts at **14 independent call sites**, each deciding for itself whether to carry the verdict-staleness fence.
- The rule governing that decision — fence when the cause is new Sentry evidence, do **not** when it is bookkeeping — lived only in prose. Over-fencing discards verdicts the pipeline means to keep; under-fencing lets a stub close over a live regression.
- PR #1716 took **seven review rounds and seven distinct defects**, every one in this area, none of them careless. Each was a *pair* of paths disagreeing about cause, so no amount of per-site care would have prevented the next one.

## The Solution

One function owns re-queueing and takes the cause as a required argument. Call sites declare *why*; they no longer reconstruct *how*.

```js
requeueQueueStub({ writeGh, readStub, readComments, claim },
                 { repo, issueNumber, cause, lastSeen, fenceProse, note,
                   dedupeFence, revalidate, onFailure, fallbackState })
```

`requeueFences(cause)` throws on anything that is not `sentry-evidence` or `bookkeeping` — **both defaults are wrong, so an undeclared cause refuses rather than guessing.** That is the difference between a rule and a convention.

## Details

- **`buildRegressedComment` has exactly one caller**, `buildRequeueFence`. A site supplies `fenceProse`, which renders *underneath* the fence line and can neither author nor omit it. A test scans every non-test `.mjs` in `scripts/` to hold that invariant, which is also what stops a fence appearing at the four bash sites below.
- **`onFailure` replaces three knobs that co-varied**: `abort` (ingest — a write failure propagates and the run retries) and `verify-end-state` (archive — reported failures are provisional, so the sequence completes and the end state is read and judged).
- **Sites 1–10 collapse into three calls.** The ingest regression reopen, the stranded-stub recovery, and the archive refusal — including `ensureSelectableForTriage`, whose reopen and label re-add moved inside the chokepoint.
- **Sites 11–14 stay in bash**, deliberately. All four are `gh issue edit` one-liners in `sentry-triage-agent.yml`, all bookkeeping, all correctly fence-free today. Moving them into node would change their `set -e` and guarded-compensation failure semantics for no safety gain; the one-caller test is what prevents a *new* fence appearing there.
- **Two contracts moved out of ingest** because the chokepoint needs them and importing ingest would cycle: the label namespace, untrusted-text neutralization, and the archive freshness-baseline fields, now in `sentry-triage-queue-contract.mjs`. That gives the archive contract its own destination without touching `project-core.mjs`, which is at 957 lines and would just become the next instance of this issue. Ingest re-exports them, so no consumer or test import path changed.

**The seven defects this makes unrepresentable**, each now a property of the chokepoint rather than of a caller: fence-iff-Sentry-evidence; one definition of the fence text; fence posted before labels with state changed last; the exclusion set recording *attempts* including decision-phase failures; "is this stub protected?" answered by running `selectVerdictComment` over live comments rather than by body presence; every body read author-fenced; live state revalidated immediately before mutating, with a failed read aborting.

## Validation

- **48 mutants, 47 killed.** The full inherited set re-derived against the new anchors — sweep S1–S9, fence P1–P5, ordering O1–O5, claim C1–C5, dedup/revalidation F2a–F3e, author fence A1–A2, admissibility R1a–R1e, note N1–N4 — plus 5 for the chokepoint: cause ignored/always-fence, cause ignored/never-fence, undeclared cause defaulting silently, a caller authoring the fence line, and the cause flipped at the archive site.
- **The single survivor is the documented redundant claim pair from #1716** (`C1`, claim-after-write). Verified against `db14fb83` that it and its twin survive there too, and that removing the catch-phase claim kills in **both** trees — so nothing drifted in this refactor. `C3` (both claims removed) and `C5` (chokepoint records successes only) both kill.
- **No test needed a real change.** The only test-file edit adds `{ cause: err }` to a new helper, for the repo's `preserve-caught-error` rule. A test that had needed a semantic change would have meant this refactor moved behaviour.
- ingest **102**, archive **113**, requeue **24** (new), digest **62**, project exit 0, autofix select 19, finalize 44, agent-comment and provider contract exit 0.
- `trunk fmt`/`check` clean · `actionlint` exit 0 · `pnpm agent:quality-gate --run --parallel 4` → exit 0.

**On file sizes:** archive 2265 → **1990**, ingest 1540 → **1180**. The new `sentry-triage-requeue.mjs` is 610 lines, just over the 600 soft cap, and roughly two-thirds of it is the rule statement in comments. It is left whole rather than splitting the rule away from the code it governs — that separation is how the rule became prose-only in the first place.

## Deferrals

- **#1698** stays open for the remaining split: Sentry-client extraction, the freshness-baseline unit, and the audit-comment builders out of `sentry-triage-archive.mjs`. That is a wide mechanical move with no correctness content — exactly the mix #1696 deferred for being hard to review alongside subtle work. Landing it here would bury a diff whose entire value is auditability.

Refs #1698, #1716, #1706, #1282

---

> [!NOTE]
> **Medium Risk**
> Changes correctness-critical triage state transitions (fencing, labels, reopen ordering) in automation that can close stubs over live regressions or discard valid verdicts if wrong; mitigated by dedicated tests, mutant suite, and unchanged bash re-queue sites.
> 
> **Overview**
> Re-queueing Sentry triage queue stubs is consolidated into **`requeueQueueStub`** in `scripts/sentry-triage-requeue.mjs`. Callers pass a required **cause** (`sentry-evidence` vs `bookkeeping`); the chokepoint alone decides whether to post the verdict-staleness regression fence, write order (fence → labels → reopen), admissibility checks via `selectVerdictComment`, and attempt tracking so a half-failed regression re-queue is not swept as fence-free bookkeeping in the same ingest run.
> 
> **`scripts/sentry-triage-queue-contract.mjs`** holds the label namespace, untrusted-text neutralization, and archive baseline fields (moved out of ingest to avoid import cycles). Ingest and archive **delegate** regression reopen, stranded-stub recovery, and live-regression archive refusal to the chokepoint; large inline helpers (`ensureSelectableForTriage`, local fence/reopen sequences) are removed from those scripts. **`pnpm sentry:requeue:test`**, runbook updates, and quality-gate wiring run requeue plus ingest/archive tests when the chokepoint or contract changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c0c0dfd0517be5882095f764d3c3b2adc7b97ac. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/notes/sentry-triage-pipeline.md
+++ b/docs/notes/sentry-triage-pipeline.md
@@ -57,6 +57,15 @@ invocations. The scripts own parsing, idempotency, and state transitions. The
 ADRs own the trust boundaries and rationale. Update those sources and this
 runbook together when a contract changes.
 
+Two shared modules sit under the stages above.
+`scripts/sentry-triage-project-core.mjs` owns the VERDICT contract — markers,
+parsing, and the two comment selectors every fence is asked through.
+`scripts/sentry-triage-queue-contract.mjs` owns the QUEUE contract — the label
+namespace, the untrusted-text neutralization, and the archive freshness-baseline
+fields. `scripts/sentry-triage-requeue.mjs` is the one place a queue stub is ever
+re-queued for triage; both producers call it and neither reconstructs the
+sequence.
+
 The triage row states an operator requirement, now backed by a mechanical
 guarantee. Every secret-bearing job in these workflows declares the
 `sentry-pipeline` GitHub Environment, whose deployment-branch policy names `main`
@@ -137,11 +146,17 @@ bulk:
   no parseable baseline;
 - other closed match: leave it closed.
 
-That write order is load-bearing at both ends. The fence goes first so no
+That write order is load-bearing at both ends, and no stage writes it by hand.
+Every re-queue in the pipeline runs through one function,
+`requeueQueueStub` in `scripts/sentry-triage-requeue.mjs`, which takes the CAUSE
+as an argument and decides the fence from it. The fence goes first so no
 interruption can leave a stub re-queued for triage without it; the state change
 goes last so none can leave it open but unselectable. Every interruption point
-then lands on a state that is inert or recoverable. The fence post is guarded by
-an exact-body check, so a retry completes the sequence without duplicating it.
+then lands on a state that is inert or recoverable. On this path the fence post
+is additionally guarded by an author-fenced identity check, so a retry completes
+the sequence without duplicating it — a guard the caller declares, and one that
+disarms itself whenever `lastSeen` does not parse, because the rendered body
+then identifies no particular occurrence.
 
 Missing or invalid timestamps fail toward re-triage. The strict timestamp gate
 prevents Sentry's long-lived regressed substatus from causing a reopen/close
@@ -166,6 +181,13 @@ Sentry loop runs, so by the time the sweep reaches a given stub the snapshot can
 be minutes old — long enough for a human to have declined it by removing the
 label, which the sweep would otherwise put straight back. A failed re-read
 leaves the stub stranded for the next run rather than recovering blind.
+
+The sweep also skips any stub the regression path ATTEMPTED this run, not merely
+the ones it re-queued. A Sentry-evidence re-queue that throws half-way would
+otherwise be recovered by the fence-free bookkeeping path seconds later, inside
+one run — a failed regression re-queue laundered into a bookkeeping one. The
+chokepoint records the attempt before its first write, so the record exists
+whether or not the writes do.
 
 The namespace is separate from the development backlog:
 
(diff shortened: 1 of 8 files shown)


chapati23:
Thanks — both P3s are fair, and the second deserves more than an acknowledgement.

**P3-2, no standalone suite for `sentry-triage-queue-contract.mjs`.** Correct, and not adding one here is deliberate: every function in it is a pure relocation, still exercised through the re-export surface, and "no test needed a real change" is the evidence the relocation was faithful. Adding a fresh suite in this PR would weaken that signal — new assertions passing tells you nothing about whether behaviour moved.

**It should not stay that way.** The follow-up split (the remaining extraction from `sentry-triage-archive.mjs`, tracked on #1698) is where the contract module stops being a pure relocation, and that is when it needs its own suite. I have noted it there rather than leaving it as a review comment nobody re-reads.

**P3-1, the 610-line file** — agreed, no action. Two-thirds of it is the rule statement, and separating the rule from the code it governs is precisely the failure this PR exists to fix: the fence rule lived only in prose across 14 sites, which is how seven distinct defects reached `main`.

One correction to point 9 for the record: CI being green on this repo says less than it looks for Sentry work. **The suites do not run in CI at all** — `.github/workflows/ci.yml` invokes no `sentry:*:test`, and `agent:quality-gate` is tested there but never run. So the ~400 assertions behind this refactor, including the 47 mutants, are enforced only by the local pre-push hook. Filed as #1721. It is not a defect in this PR, but it does mean "CI is green" and "the chokepoint's invariants hold" are independent claims here, and only the second one is backed by the evidence in the description.

PR #1775, opened 2026-08-10
fix(sentry): bind the verdict job to the round that produced its verdict

### LINKED ISSUE #1717: Sentry ingest: the stranded-stub sweep launders a regression that lands after its Sentry query
## The Problem

Ingest's stranded-stub sweep re-queues a stub as `bookkeeping`, which by design posts no verdict-staleness fence. It decides that cause from GitHub state alone, and GitHub state cannot show a Sentry occurrence that landed after this run's Sentry query. When one does, the true cause was `sentry-evidence`, the stub is re-queued fence-free, and the pre-regression verdict stays admissible.

The sequence, all in one ingest run (`runIngest`, scripts/sentry-triage-ingest.mjs):

1. `fetchMergedSentryIssues()` returns. The Sentry issue is not regressed at that moment, so the regression branch never runs for this stub and never claims it in `claimedBySentryPath`.
2. A new occurrence arrives and Sentry marks the issue regressed.
3. The sweep reaches the stub, revalidates it with `readQueueIssueState` — which requests `number,state,labels` and nothing from Sentry — and calls `requeueQueueStub` with `REQUEUE_CAUSE_BOOKKEEPING` (`recoverStrandedQueueIssue`). No fence.
4. The stub is open and selectable, so the next triage round takes it. If that round dies before posting, the `verdict` job — `if: always()` in .github/workflows/sentry-triage-agent.yml — resolves the newest admissible verdict, which is the pre-regression one, labels the stub with it, and closes it: directly, or through the serialized `project` job when the verdict is actionable and external.
5. `closedAt` now postdates that occurrence's `lastSeen`, so `decideDedupAction` returns `skip` ("closed, no events since close") until a later occurrence arrives.

The `bookkeeping` cause is right whenever the strand really was bookkeeping — a lost close response, a hand-edit, a compensation. The earlier verdict still describes the live occurrence, and fencing it would discard a good verdict. The defect is that the sweep cannot tell the two cases apart.

## What changed since filing

- **The compound archive-refusal path in the original framing is gone.** The live-regression refusal now runs through the chokepoint (`requeueQueueStub`, scripts/sentry-triage-requeue.mjs, PR #1722). A fence post that fails re-reads the stub and, when a verdict is still admissible, throws **before** the label edit — `admissibleVerdictSurvives`, which fails closed on a read error. "Fence post fails AND the label edit lands" is unreachable.
- **The fix this issue preferred has landed and does not close it.** #1722 gave re-queueing one owner with a required `cause`. That stops a caller expressing the wrong cause by accident; it cannot stop a caller whose evidence is stale. The sweep still declares `bookkeeping` truthfully as far as it knows, and still under-fences.
- **R2 survives as a footnote.** `CLOSED + needs-triage + no fence` is still reachable if someone with write access deletes the bot's fence comment between the post and the label edit and `ensureSelectableForTriage` then fails both reopen attempts. That needs a human deletion, and the run is already RED with `::error:: … STRANDED … Reopen it by hand`.

## Preferred fix: bind the verdict job to a verdict the round actually produced

Kill the consequence instead of inferring the cause. The hazard is that `always()` lets the `verdict` job act on a verdict comment from an earlier round. Bind it to this one:

- `select` records, per stub, the id of the newest trusted verdict comment visible before triage runs, and emits them as a job output. A failed read fails closed for that stub.
- `verdict` passes that id to `scripts/sentry-triage-project.mjs --parse-only`, and `resolveVerdict` (scripts/sentry-triage-project-core.mjs) refuses when the comment it selects carries that same id — the round posted nothing new.

The plumbing already exists: `selectVerdictComment` returns the comment `url`, `resolveVerdict` passes it out as `verdictCommentUrl`, and `verdictCommentIdFromUrl` parses the numeric id — all built as the autofix generation token for #1506 and consumed today by scripts/sentry-autofix-select.mjs. Comment ids are unique and never reused, so there is no window to race. This closes the path whatever the cause was, adds no Sentry call to the sweep, and leaves the fence in place as defense in depth for the digest, projection and autofix consumers.

## Rejected

- **A per-stub Sentry lookup in the sweep.** A short-ID lookup plus an issue fetch for every stranded stub, introducing Sentry coupling into a sweep that needs none — and still racy, because any Sentry read has a window after it.
- **Fencing the recovery unconditionally.** It discards a valid verdict on every legitimate bookkeeping recovery. Over-fencing is exactly as wrong as under-fencing, and quieter.
- **Persisting the cause on the stub** — a body marker or a dedicated label written before the label edit. A durable write on a path that is already partially failed, and a second source of truth for the cause.

## Done means

The `verdict` job cannot label or close on a verdict comment that predates the triage round it is settling. Proven by a test driving select → a triage round that posts nothing → verdict, asserting the job refuses and leaves `sentry:needs-triage` for the next run, plus a mutant that drops the binding and must fail.

The sweep's blind spot is also written up in docs/notes/sentry-triage-pipeline.md beside the existing known residual. The binding kills the consequence, not the blind spot, and the blind spot is documented nowhere today.

## History

The original framing was a compound failure in the archive's live-regression refusal, laundered by a later sweep. It required, in one archive run: the best-effort fence post failing, the label edit landing, and `ensureSelectableForTriage` then failing to reopen — plus `lastSeen <= closedAt` for the sweep to reach the stub at all. #1716 left it open on purpose: two independent writes have to fail together, the run is already RED when they do, and redesigning the producer/recovery boundary inside a PR that had already taken three P1s was the wrong trade. #1716 then moved the abort ahead of the label edit; #1722 made that ordering a property of the chokepoint rather than of the call site. The archive-side analysis is kept in the comments below.

Refs #1706, #1716, #1722, #1698, #1696, #1282

chapati23:
## The Problem

- Ingest's stranded-stub sweep re-queues with cause `bookkeeping` and posts no staleness fence. It decides that from GitHub state, and GitHub state cannot show a Sentry occurrence that landed after this run's `fetchMergedSentryIssues()` returned.
- The `verdict` job runs `if: always()`, so a triage round that dies before posting still reaches it — and settles the stub on the previous round's verdict, which describes the pre-regression occurrence.
- The close then postdates that occurrence's `last_seen`, so ingest's own dedup gate skips the stub until the next one: a live regression laundered into a settled ledger entry.

## The Solution

Kill the consequence instead of inferring the cause. `select` records, per stub, the id of the verdict comment already on it; the `verdict` job passes that token to `--parse-only`, and `resolveVerdict` refuses unless the comment it selects is strictly newer. A round that posted nothing cannot settle the stub, whatever the re-queue's cause was — and the sweep gains no Sentry call.

## Details

- **Where the binding lives.** `resolveVerdict` takes an optional `priorVerdictCommentId`. Only `--parse-only` passes one; projection and autofix-select run downstream of a verdict the binding already cleared and keep the pre-#1717 behaviour.
- **Why `select`.** It is trusted and it runs before the agent, so what it reads is what the round started from. Recorded any later, the token would include the round's own comment and prove nothing. A workflow-shape test refuses a second recorder anywhere else in the file.
- **One selector, both ends.** The recorder selects through `selectVerdictComment`, the same selector the settlement resolves with, so the two cannot drift into disagreeing about what "the previous verdict" is.
- **Strictly newer, not merely different.** Refusing on the same id is the common case: the round posted nothing. Refusing on an *older* id covers the recorded comment being deleted between the two reads — the same write-access premise the surviving archive-side route needs — which an equality check would wave through onto an even older verdict. GitHub issue-comment ids come from one monotonically increasing sequence, so "newer" is a numeric comparison.
- **Fail closed per stub, never per batch.** An unreadable stub, or a verdict comment with no parseable `#issuecomment-<n>` id, records `unknown`, which never passes. That costs one loud triage round on that stub and self-heals: the round's own verdict comment becomes the next run's baseline. One unreadable stub does not starve the rest of the batch. The refusal lands before the label edit, so the stub keeps `sentry:needs-triage`.
- **Reused plumbing.** The token is the `#issuecomment-<n>` generation token the autofix leg already threads for its ABA check (#1506). The fence is untouched and still decides admissibility for the digest, projection and autofix consumers; the binding decides whether a run may settle the stub at all.
- **The sweep is unchanged, deliberately.** A per-stub Sentry lookup and unconditional fencing were both rejected in the issue — the first couples a sweep that needs no Sentry and is still racy, the second discards a good verdict on every genuine bookkeeping strand. The sweep's blind spot is now written up beside the existing residual in the runbook, which recorded only the within-a-run laundering.
- **Not an ADR.** This refines a mechanism inside ADR 0036/0038's existing decision rather than making a new one; `pnpm adr:check` agrees. The rationale is recorded in the runbook's new "The round binding" section and at each site.
- **Composes with what just landed.** #1764 owns the verdict step's shed list and its post-condition compensation — untouched; the binding refuses before that window opens. #1766's re-queue split and baseline work is upstream of the sweep and unaffected.

## Validation

- `bash scripts/agent-quality-gate.sh --run --parallel 4 --base origin/main` — all 13 mapped commands pass (includes `lint:scripts`, `sentry:project:test`, `sentry:ingest:test`, `sentry-triage-agent-comment.test.mjs`, trunk check on all seven touched files).
- `pnpm agent:autoreview --engine local` — clean, no findings.
- Eleven mutants, each killed by a named test: dropping the binding from `resolveVerdict`; letting `unknown`, an equal id, or an older id pass; recording `none` on a failed read; unthreading the token at either end; accepting an out-of-set token; and three workflow-shape mutants (verdict step stops passing the token, select stops emitting the record, a missing record degrades to empty instead of `unknown`).
- Two of those tests reproduce the laundering end to end against the **real** sweep and the **real** resolver: a regression landing after the Sentry query, and the surviving archive-refusal route whose fence was deleted. Both fail without the binding with `expected the resolution to refuse`.
- A third scenario test pins the other direction — a swept stub whose next round *does* post a verdict still settles normally, so the binding is not a blanket refusal.
- The independent-model review comes from this PR's external reviewers; the codex CLI is not on PATH in this environment.

## Deferrals

- None

Closes #1717

---

> [!NOTE]
> **Medium Risk**
> Changes deterministic queue settlement in CI for all triage runs; mis-wiring could block legitimate closes, but failures keep `sentry:needs-triage` and retry.
> 
> **Overview**
> Fixes a path where ingest’s **fence-free bookkeeping** stranded-stub recovery could leave an old verdict admissible, then the always-on **`verdict` job** would re-label and close the stub—burying a Sentry regression that arrived after ingest’s query.
> 
> **Round binding:** Before the agent runs, **`select`** records each stub’s existing verdict comment id via `--prior-verdicts` (GitHub-only). **`verdict`** passes that token to `--parse-only`; **`resolveVerdict`** / `priorVerdictRefusal` only allow settlement when the selected verdict comment is **strictly newer** than the baseline (`none` / numeric id / fail-closed `unknown`). The stranded sweep is unchanged; this closes the consequence at settlement.
> 
> Docs add **“The round binding”** and operator guidance for intentional refusals. Tests cover parser/workflow shape and end-to-end sweep → settle scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14148b237d70dee262c0d297396db7215c3dcc4e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/sentry-triage-agent.yml
+++ b/.github/workflows/sentry-triage-agent.yml
@@ -51,6 +51,12 @@ jobs:
   # and the secret guard live inside the first step, so a disabled or
   # unprovisioned state is a clean no-op (`[]`, exit 0), never a failure and
   # never an empty matrix expression downstream.
+  #
+  # It also records, per selected stub, the verdict comment already on it — the
+  # round binding for the `verdict` job below (issue #1717). This is the only
+  # job that can: it is trusted AND it runs before the agent, so what it reads
+  # is what the round started from. The read is GitHub-only; no Sentry call and
+  # no Sentry credential enter this job.
   # ---------------------------------------------------------------------------
   select:
     runs-on: ubuntu-latest
@@ -68,7 +74,22 @@ jobs:
     outputs:
       issues: ${{ steps.select.outputs.issues }}
       count: ${{ steps.select.outputs.count }}
+      priorVerdicts: ${{ steps.prior-verdicts.outputs.priorVerdicts }}
     steps:
+      # SHA-pinned (not tags) so tag retargeting / a compromised release cannot
+      # silently run new code. Bump via Dependabot (.github/dependabot.yml).
+      # This checkout is PRISTINE — no agent runs in this job — so the
+      # prior-verdict recorder below is safe to run from the checkout path, the
+      # same reasoning the `verdict` job's plain `scripts/` invocation rests on.
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 1
+          persist-credentials: false
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version-file: .node-version
+
       - name: Select queue issues (kill-switch + secret guarded)
         id: select
         env:
@@ -136,6 +157,57 @@ jobs:
             echo "count=${count}"
           } >> "${GITHUB_OUTPUT}"
           echo "Selected ${count} issue(s): ${issues}"
+
+      # -----------------------------------------------------------------------
+      # Round binding, recorded BEFORE the agent runs (issue #1717).
+      #
+      # The `verdict` job is `if: always()`, so a triage round that dies before
+      # posting still reaches it, and it settles the stub on the newest
+      # ADMISSIBLE verdict comment. The regression fence only makes a verdict
+      # inadmissible when some producer posted a fence; ingest's stranded-stub
+      # sweep re-queues for BOOKKEEPING and posts none, by design — nothing
+      # about the Sentry issue changed as far as it can see. It reads GitHub
+      # only, so a Sentry occurrence that landed after its own
+      # `fetchMergedSentryIssues()` returned is invisible to it, and the
+      # previous round's verdict — which describes the pre-regression
+      # occurrence — stays admissible. Without this binding the `verdict` job
+      # then labels and closes on it, and the close postdates the occurrence, so
+      # ingest's dedup gate skips the stub until the NEXT occurrence: a live
+      # regression laundered into a settled ledger entry.
+      #
+      # So record what each stub already carried and make the verdict job prove
+      # the round moved it forward. This is the same generation-token plumbing
+      # the autofix leg threads (issue #1506), and it kills the consequence
+      # whatever the re-queue's cause was — no per-stub Sentry lookup is added
+      # to the sweep, which is what makes it a design fix rather than a guard.
+      #
+      # `[]` (kill switch off, secrets unprovisioned, or an empty queue) emits
+      # `{}` and costs nothing. A per-stub read failure records `unknown`, which
+      # the verdict job refuses — fail closed for that stub only, never a
+      # starved batch.
+      # -----------------------------------------------------------------------
+      - name: Record the verdict comment each stub arrives with
+        id: prior-verdicts
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SELECTED_ISSUES: ${{ steps.select.outputs.issues }}
+        run: |
+          set -euo pipefail
+          # Plain scripts/ is correct here for the same reason it is in the
+          # verdict job and nowhere near the agent: this job has its own runner
+          # and its own pristine checkout, which no agent ever touched.
+          prior_file="${RUNNER_TEMP}/prior-verdicts.json"
+          node scripts/sentry-triage-project.mjs \
+            --prior-verdicts \
+            --issues "${SELECTED_ISSUES}" \
+            --repo "${GITHUB_REPOSITORY}" \
+            > "${prior_file}"
+          # Compact to one line: GITHUB_OUTPUT is line-oriented, and the script
+          # already prints a single line — re-compacting is the cheap assertion
+          # that it stayed that way.
+          prior=$(jq -c . "${prior_file}")
+          echo "priorVerdicts=${prior}" >> "${GITHUB_OUTPUT}"
+          echo "Prior verdict comments: ${prior}"
 
   # ---------------------------------------------------------------------------
   # triage: one claude-code-action run per queue issue (max 2 in parallel).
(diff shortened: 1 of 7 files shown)


chapati23:
Thanks for the thorough read-through.

On the one P3 — the prior-verdicts step running checkout/setup-node even when `select.outputs.issues == '[]'`: leaving it as-is. It's a correct no-op, the cost is ~20s on a twice-daily scheduled run, and gating it now means a fresh commit that resets the current green Codex approval into another full review cycle. That trade isn't worth it for the saving. Noted as a free ride-along the next time this workflow is edited for another reason.

Issue #1789, opened 2026-08-11
fix(terraform): restore Aegis App Engine staging access


chapati23:
## Problem

Aegis App Engine deployments fail before version creation because the App Engine default service account cannot write to the Google-managed staging bucket:

`Unable to write to staging bucket staging.mento-monitoring.appspot.com`

The failure predates PR #1776 and has recurred since July 30. Production still serves the previous version, but every new Aegis deployment is blocked.

## Recovery

- Model the exact staging-bucket IAM grant in the platform Terraform stack.
- Keep access scoped to `staging.mento-monitoring.appspot.com` and `mento-monitoring@appspot.gserviceaccount.com`.
- Preserve Google-managed bucket IAM with non-authoritative member resources.
- Add fail-closed contract coverage and update ADR/runbook claims.
- Review a clean current-main plan; apply only after separate explicit approval.
- Rerun Aegis deployment and verify the promoted version plus live `/metrics` freshness.

## Evidence

- Failed run: https://github.com/mento-protocol/monitoring-monorepo/actions/runs/31470858099
- Same failure: https://github.com/mento-protocol/monitoring-monorepo/actions/runs/30801805682
- Last successful Aegis deploy: https://github.com/mento-protocol/monitoring-monorepo/actions/runs/30369655615

chapati23:
Agent claim: codex claimed #1789 for implementation.

Claim ID: codex-20260811T105653-f0bzfn
Branch: fix/aegis-app-engine-staging-iam
Claimed at: 2026-08-11T10:56:53.988Z

chapati23:
Moved to review: #1789 is now represented by PR #1793.
Branch: fix/aegis-app-engine-staging-iam

chapati23:
Recovery remains incomplete. PR #1793 added bucket-scoped Storage Admin for AppSpot, and a clean current-main platform plan on 2026-08-13 confirmed that grant is live. Rerun attempt 2 of Actions run 31675490741 still failed with the same staging-bucket denial. The App Engine CreateVersion audit event identifies metrics-bridge-deployer@mento-monitoring.iam.gserviceaccount.com as the caller. The follow-up must grant the modeled App Engine uploader callers access to only staging.mento-monitoring.appspot.com, apply after separate approval, then rerun and verify the live service.

chapati23:
Agent claim: codex claimed #1789 for implementation.

Claim ID: codex-20260813T142618-pr3bi0
[...]

PR #1837, opened 2026-08-13
fix(terraform): restore Aegis staging uploads

chapati23:
## The Problem

- Aegis App Engine deployments fail because the `gcloud` caller cannot write to `staging.mento-monitoring.appspot.com`.
- The existing bucket-scoped grant covers the default AppSpot service account, but the failed `CreateVersion` audit event names `metrics-bridge-deployer` as the caller.

## The Solution

- Give the existing App Engine uploader set Storage Admin on only the service-owned App Engine staging bucket so version submission can complete without restoring project-wide storage access.

## Details

- Adds a bucket IAM grant for `local.app_engine_source_uploaders` on `staging.${google_project.monitoring.project_id}.appspot.com`.
- Retains the separate default AppSpot grant required after App Engine takes over the deployment.
- Extends the deploy-staging and production-infrastructure identity contracts so the caller set, bucket, role, member expression, and App Engine dependency remain pinned.
- Updates ADR 0053 and the live Terraform, deployment, checklist, and agent guidance to reflect the two-principal path.
- Risk is limited to stronger object and bucket authority for the existing uploader set on one service-owned staging bucket. No project-wide grant or Terraform-managed source-bucket grant is added.

Refs #1789

## Validation

- `pnpm agent:quality-gate --run` — passed all mapped checks, including Terraform init/validate, `pnpm tf:test`, script lint, docs checks, and targeted Trunk checks.
- `pnpm agent:autoreview --engine local` — clean.
- Fresh-context semantic autoreview — no findings; patch correct at 0.94 confidence.
- `git diff --check` — passed.

## Deferrals

- https://github.com/mento-protocol/monitoring-monorepo/issues/1789 — apply the platform Terraform change after explicit approval, rerun the Aegis workflow, and record live deployment proof.
- https://github.com/mento-protocol/monitoring-monorepo/issues/1839 — update the bounded feedback parser for Claude's new explicit LGTM/no-action summary shape; this is separate agent tooling and does not change the Aegis IAM fix.

## Checklist

- [x] Architecture decision? ADR 0053 is updated because it owns this deployment-staging permission split.


chapati23:
Won't fix: The bucket-scoped `roles/storage.admin` grant intentionally applies to `local.app_engine_source_uploaders` because every identity in that set is a supported `gcloud app deploy` caller. Limiting the grant to the one caller observed in this failure would leave the manual developer and Alloy-backed deployment paths unable to use App Engine's service-owned staging bucket. The authority remains limited to `staging.<project>.appspot.com`, and ADR 0053 records this trade-off.

Disposition for Claude's P3 note: https://github.com/mento-protocol/monitoring-monorepo/pull/1837#issuecomment-5281908631

PR #1850, opened 2026-08-14
fix(dashboard): bind browser fixtures to source identity

### LINKED ISSUE #1465: ui-dashboard: run-browser-tests.mjs skips fixture rebuild on stale direct (non-turbo) re-runs
## Problem

`ui-dashboard/scripts/run-browser-tests.mjs` skips `buildFixtureApp()` whenever `.next-fixture/BUILD_ID` already exists (`fixtureBuildExists()`). When invoked through `turbo run test:browser`, this is safe: turbo's own input hashing already reran the `fixture-build` task if any tracked input changed before `test:browser` starts. But direct callers that bypass turbo — `pnpm test:browser`, `pnpm test:browser:update-snapshots`, or a bare local run — have no such guarantee: running one of these twice with a source edit in between silently serves the stale bundle from the first run, with no signal to the developer that their edit wasn't picked up.

Flagged by Codex on PR #1456 (`ui-dashboard/scripts/run-browser-tests.mjs:95`, review comment https://github.com/mento-protocol/monitoring-monorepo/pull/1456#discussion_r3630084719).

This is a real gap for anyone iterating on browser tests/snapshots directly (bypassing turbo) but does not affect CI (fresh checkout every run, `.next-fixture` never present) or the turbo-invoked gate path (turbo's caching already handles it).

## Solution ideas

- Persist a lightweight content signature (e.g. mtime+size or a hash) of the same source set turbo's `fixture-build` task tracks (see `turbo.json`'s `fixture-build.inputs`) alongside `BUILD_ID`, and compare it before skipping the build. Keep the tracked-file list in one place (e.g. export it from `fixture-constants.mjs` and read it from `turbo.json` generation, or vice versa) to avoid the exact "two places to update" hazard PR #1456 fixed for the port/dist-dir constants.
- Alternative: route the direct/bypass-turbo entry points through `turbo run fixture-build` internally instead of reimplementing staleness detection in JS.

### LINKED ISSUE #1466: ui-dashboard: fixture Hasura server reuse can't prove it's the current checkout's fixtures
## Problem

`playwright.config.ts`'s fixture-server `webServer` entry uses `reuseExistingServer: true` on the fixed port `3211`, on the premise that "a healthy [server] left over from a prior run (or a sibling worktree) is safe to reuse." The `/health` endpoint (`tests/browser/fixtures/hasura-fixture-server.mjs`) only reports `{ ok, counters }` — nothing that proves the responding server's fixture data/schema matches *this* checkout's `tests/browser/fixtures/hasura-fixture-server.mjs`. If that file changes between the two runs/worktrees sharing the port, tests silently run against stale or incompatible fixture responses instead of failing loudly.

Flagged independently by two reviewers on PR #1456: Codex (inline, https://github.com/mento-protocol/monitoring-monorepo/pull/1456#discussion_r3630084725) and claude[bot]'s top-level review ("flagging for visibility only in case reviewers want a second opinion on it").

This is a deliberate, documented tradeoff (ADR 0025 addendum, opt-out via `PLAYWRIGHT_REUSE_FIXTURE_SERVER=false`), but the specific correctness gap — no identity check before reuse — was not addressed, only the flakiness tradeoff it replaces.

## Solution ideas

- Add an identity marker to the fixture server's `/health` response (e.g. a hash of the fixture server's own source, or a monotonic build/version stamp written by `fixture-build.mjs`), and have the reuse decision compare it before treating an already-running server as reusable, failing over to a restart when it doesn't match.
- Playwright's built-in `webServer.url` health check only polls for HTTP 200; a custom pre-check (e.g. a `globalSetup` step) would be needed to inspect the body before deciding whether reuse is safe.

chapati23:
## The Problem

- Direct browser-test runs trusted any existing `.next-fixture/BUILD_ID`, so source edits could silently reuse a stale production bundle outside Turbo.
- Playwright reused any healthy process on fixed port 3211 without proving that its fixture data and runtime mode matched the current checkout.

## The Solution

- Bind cached fixture builds to Turbo's authoritative `fixture-build` hash and bind fixed-port server reuse to a fail-closed identity over its behavior sources, scenario, and delay.

## Details

- Writes the build identity only after a successful build and invalidates the old marker before rebuilding.
- Qualifies health checks with the expected server identity; mismatched or unverifiable processes are left untouched and rejected.
- Shares runtime option normalization between the preflight and server, including the pool Lighthouse scenario.
- Makes the package runner the supported Playwright entry point and updates Turbo inputs, environment tracking, the ADR, and the dashboard verification runbook.

## Validation

- Focused fixture identity, endpoint, and runner suite — 39/39 tests passed.
- Real package runner CLI with `--grep 'filters homepage pools by name'` — 1/1 Playwright test passed.
- Dashboard typecheck, lint, coverage, and the full 27-test Playwright browser suite — all passed.
- Playwright config contract — lists all 27 tests with a valid identity and fails closed without one.
- `CI=true pnpm agent:quality-gate --run` — all mapped commands passed.
- Fresh independent semantic review of immutable bundle `d58561748c4a546769a874c659082472360112433f407ebe5ca4714464cb95d6` — clean at 0.96 confidence; deterministic local autoreview was also clean.
- Chrome DevTools browser verification was attempted, but the shared MCP profile was locked by an unowned browser process; Playwright supplied the real-browser proof without stopping that process.

## Deferrals

- None

## Checklist

- [x] Architecture decision? The existing fixture-browser ADR is updated with the identity invariant.

Closes #1465
Closes #1466

---

> [!NOTE]
> **Medium Risk**
> Changes affect CI/local browser gates and fixed-port reuse behavior; logic is heavily tested but misconfiguration could block runs until stale servers are stopped manually.
> 
> **Overview**
> **Fixture browser runs** no longer trust a bare `.next-fixture/BUILD_ID` or any healthy process on port **3211**. Direct runs compare cached output to Turbo’s current **`fixture-build`** hash (via new **`fixture-identity.mjs`**), invalidate before rebuild, and write identity only after a successful build. The runner probes **`/health/{identity}`**; Playwright requires **`PLAYWRIGHT_FIXTURE_SERVER_IDENTITY`** and reuses the fixed-port server only when preflight sets **`PLAYWRIGHT_REUSE_FIXTURE_SERVER=true`**.
> 
> The **Hasura fixture server** exposes identity on qualified health (409 on mismatch), shares scenario/delay normalization with Lighthouse, and keeps unqualified **`/health`** for non-Playwright callers. **`run-browser-tests.mjs`** is refactored for testability; Turbo inputs/env, quality-gate assertions, ADR 0025, and the dashboard verification runbook are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16028f473e44cadae46dd7d71f06e0487453f9aa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chapati23:
Fixed in b5463ed7 — Updated the runner header to state that only identity-matching fixture servers are reused and mismatched or unverifiable processes fail closed.

chapati23:
@codex review

Issue #1831, opened 2026-08-13
Peg deviation history: Grafana read token (IaC), ADR 0049 amendment, history route + chart wiring


chapati23:
## Context

The 2a peg-monitoring board (see the board-redesign issue) needs a 24h/7d/30d
peg-deviation history series per asset for its Peg History chart. The series
exists only in Grafana Cloud Mimir (`mento_peg_deviation_bps`, scraped from
metrics-bridge via Alloy; verified live with data back to 2026-07-29).
metrics-bridge deliberately keeps no history and ADR 0049 rejected both
"persist history in the bridge" and "query Grafana from the dashboard" — for
current-state evidence. The owner has approved reading **history** from
Grafana Cloud with a dedicated read-only token, recorded as an ADR amendment:
current-state stays on the decision-package read model; history reads go to
the authoritative metrics store with an isolated failure domain.

## Scope

1. **Terraform (this is the review-critical part; put it up early):** mint a
   least-privilege Grafana Cloud service-account token (metrics read + access
   to the alert state-history API for the follow-up alerts-feed issue; NOT the
   Admin token used by `alerts/rules`), modeled fully in IaC per ADR 0030, and
   deliver stack URL + token to the dashboard's Vercel project env
   (server-side only, never `NEXT_PUBLIC_*`). Respect stack ownership in
   `terraform.stacks.json`; plan-only on PR, human-approved apply.
2. **ADR:** amend/supersede ADR 0049 with the history-read boundary.
3. **Dashboard route (after token lands):** `/api/peg-monitoring/history` —
   PromQL range query per range (24h/5m, 7d/30m, 30d/2h steps) for the deep
   source per asset, Zod-validated, bounded response, `no-store`; failure
   isolated so the board never degrades when history is down.
4. Wire the board's chart + range chips to the route; handle series that start
   mid-window (data exists only from 2026-07-29) without extrapolating.

## Done means

- `terraform plan` output on the PR showing only the new SA/token/env
  resources; no apply without human approval.
- ADR merged in the same PR as the terraform.
- Route returns validated series for all three ranges in production; chart
  renders them; board unaffected when the route 5xxs.

chapati23:
Agent claim: claude claimed #1831 for implementation.

Claim ID: claude-20260813T120443-8cyi2l
Branch: feat/peg-history-grafana-token
Claimed at: 2026-08-13T12:04:43.748Z

chapati23:
[...]

PR #1852, opened 2026-08-14
fix: align peg history queries to Grafana steps

chapati23:
## The Problem

- Live proof after the Grafana credential rollout returned HTTP 502 for the `24h` and `7d` peg-history ranges, while `30d` succeeded.
- Grafana aligns Prometheus range-query bounds to the selected step, but the route rejected those valid boundary points against the unaligned browser timestamp.

## The Solution

- Floor the query end to the selected range step, then derive the start from that same bound so the request, parser, and response share one exact window.
- Run every range test with an unaligned clock and cover retained-package timestamps so the production failure cannot hide behind step-aligned fixtures again.

### Invariants

- The requested asset, deep source, and policy version remain exact selectors.
- Each range keeps its reviewed window, step, point cap, timeout, and response-size limit.
- The aligned query end never exceeds the live request time or retained package timestamp.
- Grafana history remains isolated from the current-state decision package and board.

### Degraded behavior

- Grafana errors and invalid frames still fail closed through the existing history-unavailable state.
- A history failure still cannot block or replace the current-state peg board.

### Intentional non-goals

- No credential, datasource, PromQL selector, range, chart, or current-state behavior changes.
- Production API and browser proof remain a post-merge redeploy step for issue #1831.

## Validation

- Live reproduction before the fix: unaligned `24h` and `7d` requests returned 502; step-aligned requests returned 200; `30d` returned 200 with 196 points.
- `pnpm --filter @mento-protocol/ui-dashboard test -- src/app/api/peg-monitoring/history/__tests__/route.test.ts` — 296 files and 4,280 tests passed.
- `pnpm agent:quality-gate --run` — all mapped commands passed, including production browser fixtures, coverage, lint, typecheck, Trunk, knip, dependency checks, React Doctor, bundle size, Terraform contracts, and documentation checks.
- Exact post-commit gate for `30ff0dd2` passed with the same mapped command set.
- Local autoreview deterministic guard: clean.
- Fresh-context Codex semantic autoreview: clean, no findings.
- Claude Security scan: skipped because the plugin is unavailable in Codex.

## Deferrals

- None

## Checklist

- [x] Architecture decision? Not applicable; this PR preserves ADR 0063 and updates its applied-state note.

Follow-up for #1831.

---

> [!NOTE]
> **Low Risk**
> Narrow bugfix to time-window alignment on an isolated history route; selectors, ranges, and fail-closed behavior unchanged.
> 
> **Overview**
> Fixes **502s on `24h`/`7d` peg history** after Grafana rollout by matching Grafana’s step-aligned Prometheus window instead of using raw browser time.
> 
> The history route now **floors the query end** (`to`) to the selected range’s step, derives `from` from that bound, and uses the same window for the upstream request, response metadata, and point validation so step-aligned boundary samples are not rejected as out of range. **ADR 0063** is updated to record that the Grafana read identity Terraform was applied on 2026-08-14.
> 
> Tests use an **unaligned clock** and assert aligned `from`/`to` for all ranges and retained-package `to` timestamps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30ff0dd2ddfc635ad9b94d21d1fa8e81fc5e64f0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/adr/0063-dashboard-grafana-history-read-access.md
+++ b/docs/adr/0063-dashboard-grafana-history-read-access.md
@@ -15,9 +15,9 @@ garden_lane: adrs-architecture
 
 **Status:** Accepted (Aug 2026). Amends
 [ADR 0049](0049-peg-decision-package-read-model.md), which stays in force for
-current-state evidence. The Terraform is checked in and unapplied; a
-human-approved platform apply activates the identity and the dashboard
-environment. **Scope:** ui-dashboard / terraform/infra
+current-state evidence. The Terraform was applied on 2026-08-14, and the
+dashboard environment now uses the dedicated read identity. **Scope:**
+ui-dashboard / terraform/infra
 
 ## Context
 
--- a/ui-dashboard/src/app/api/peg-monitoring/history/__tests__/route.test.ts
+++ b/ui-dashboard/src/app/api/peg-monitoring/history/__tests__/route.test.ts
@@ -8,7 +8,7 @@ import {
   resolveGrafanaQueryEndpoint,
 } from "../route";
 
-const NOW_MS = Date.UTC(2026, 7, 13, 20, 0);
+const NOW_MS = Date.UTC(2026, 7, 13, 20, 17);
 const baseQuery = {
   asset: "europ-schuman",
   source: "bitvavo_eur",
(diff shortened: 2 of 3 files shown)


chapati23:
@claude review the current head again. Keep the review substantive and report any actionable concern normally. For the repository feedback projector, use this exact top-level structure with no prose between the verdict and Findings:

### Review: fix: align peg history queries to Grafana steps

**Verdict: LGTM**

### Findings

If and only if the review is clean, use this exact finding line:
1. [P3] None blocking: clean, well scoped fix. Tests cover the changed paths.

### Roll-up

If and only if the review is clean, use this exact roll-up line:
1. [P3] No-action: fix is correct.

If you find an actionable concern, replace the clean verdict and lines with the appropriate P1/P2/P3 finding instead of forcing LGTM.

PR #1857, opened 2026-08-14
fix(alerts): schedule on-call reconciliation after handover

gisk0:
## The Problem

- On-call reconciliation polls every 15 minutes despite the observed Sunday 10:00 Europe/Berlin handover.
- A scheduler-only change currently risks an unnecessary function redeploy.

## The Solution

- Check at 10:05 and 10:20 each Sunday, shortly after handover, without changing the function runtime configuration.

## Details

- Update the Terraform defaults, operator documentation, and the production-infrastructure provenance hash.
- Removing the schedule from the runtime hash updates Cloud Scheduler without redeploying Cloud Run or changing invocation access.

## Validation

- CI Terraform plan, registry validation, cloud-function quality checks, production-infrastructure contract, and Sentry suites passed.
- The root-script suite is rerunning after its previous 15-minute runner cutoff.

## Deferrals

- None

## Checklist

- [x] Architecture decision: not applicable.

### DIFF
--- a/alerts/infra/README.md
+++ b/alerts/infra/README.md
@@ -176,7 +176,7 @@ chain key.
 
 ### On-call Announcer
 
-- Runs from Cloud Scheduler every 15 minutes by default
+- Runs from Cloud Scheduler at 10:05 and 10:20 every Sunday in `Europe/Berlin`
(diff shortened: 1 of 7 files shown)


gisk0:
@codex review

Issue #1877, opened 2026-08-17
Scripts reorg: restructure scripts/ into module directories (P0-P13)


chapati23:
## Problem

`scripts/` holds 233 files and 160,736 lines. 210 files sit flat in one directory. Consumer paths are pinned across six systems: 73 root package.json entries, 22 workflow files, .trunk/trunk.yaml, .claude/settings.json plus its verbatim copy in check-agent-context.mjs, the pre-install alias registry, and docs/notes/quick-commands.md. Silent-failure hazards guard any move: two `scripts/deploy-*.sh` globs stop matching after a move, check-deploy-root-anchors.test.mjs passes vacuously on an empty subject list, and a missed implementation_signature() entry makes `--skip-if-fresh` skip real pre-push work.

## Plan

Operator-approved plan from the 2026-08-17 investigation: 14 PRs now (P0–P13) plus 5 deferred tracks (D1–D5) behind named triggers. 89 files move into eleven `scripts/` subdirectories plus a shared `scripts/lib/` tier. 49 files stay flat; each has a named pin. Seven large files split. Four bash scripts become Node. One dead file is deleted. No workspace packages.

Two lanes run after P1: a serial metadata spine (P3 context → P4 docs → P5 pr) and a parallel hygiene/infra lane (P2, P6, P7→P8, P9, P10). P11–P13 close the programme. Every phase updates scripts/AGENTS.md and its slice of the six-system sweep in the same PR. Each PR needs the Sentry suites green before merge because two suites declare readsDirs ['scripts'].

## Phases

- [x] P0 groundwork: delete code-health-history.test.mjs, harden check-deploy-root-anchors.test.mjs (recurse + assert orderedAnchors coverage + fail on empty set), module-directory ADR, scripts/AGENTS.md directory map and pin registry
- [x] P1 pilot: scripts/workflows/ (9 files) + check-autofix-ci-trust split
- [x] P2 scripts/bootstrap/ (5 files) + lib/install-marker.sh
- [x] P3 scripts/context/ (9 files) + lib/gh-issue-lifecycle.mjs
- [x] P4 scripts/docs/ (10 files); update the gate dynamic import and implementation_signature()
- [x] P5 scripts/pr/ (11 files) + agent-issue-board 5-way split; fix the pr-description.yml trusted-base probe
- [x] P6 scripts/supply-chain/ + scripts/mcp/ (18 files) + lib/pnpm-override-selector.mjs
[...]

PR #1891, opened 2026-08-17
refactor(scripts): extract generic hcl and workflow-yaml cores into scripts/lib/ (P7) (security review)

chapati23:
## The Problem

- Two generic parsing cores lived inside `scripts/production-infra-identity-contract/`: an HCL tokenizer and a set of Actions-workflow helpers. Neither holds identity or IAM logic, and five files outside that directory already imported them — including the two ADR 0053 deploy-staging contract files.
- A directory named for one domain owning a parser five outside files depend on invites the next domain to write its own copy.
- `peg-policy.mjs` had grown to 1,348 lines covering four separate concerns.

## The Solution

- `hcl.mjs` moves to `scripts/lib/hcl.mjs` unchanged. The six generic workflow-YAML helpers become `scripts/lib/workflow-yaml.mjs`. Domain policy — inventories, pinned hashes, expected identities — stays where it was.
- `peg-policy.mjs` splits into four modules by concern. `index.mjs` remains the single facade.
- Every moved line is byte-identical to its source. The tables below give the command to prove each one.

## Details

Phase P7 of the `scripts/` reorganization, tracked at <https://github.com/mento-protocol/monitoring-monorepo/issues/1877>.

### Byte-equality proof

`scripts/lib/hcl.mjs` is a pure `git mv`. The commit records `rename scripts/{production-infra-identity-contract => lib}/hcl.mjs (100%)`, and `git diff -M --stat` reports zero insertions and zero deletions for it. For the split files, each moved region is verbatim. Run any row:

```
diff <(git show origin/main:<source> | sed -n '<from>,<to>p') \
     <(sed -n '<from>,<to>p' <target>)
```

| Source (on `origin/main`) | Target | Lines |
| --- | --- | --- |
| `production-infra-identity-contract/workflow-inventory.mjs` L195-263 | `lib/workflow-yaml.mjs` L6-74 | 69 |
| `production-infra-identity-contract/peg-policy.mjs` L18-108 | `peg-policy-constants.mjs` L1-91 | 91 |
| `production-infra-identity-contract/peg-policy.mjs` L110-532 | `peg-policy-bucket.mjs` L29-451 | 423 |
| `production-infra-identity-contract/peg-policy.mjs` L533-954 | `peg-policy-publication.mjs` L15-436 | 422 |
| `production-infra-identity-contract/peg-policy.mjs` L955-1348 | `peg-policy-runtime.mjs` L34-427 | 394 |

The only `peg-policy.mjs` lines that did not carry forward are its original 16-line `./hcl.mjs` import header and two blank separators (L17, L109). Every other line moved untouched. The extraction ran from a script that anchors each seam on the exact source line and asserts byte-equality; a source edit fails the script rather than producing a wrong split.

### What the split changed semantically

Splitting a module widens its surface: helpers that were file-private are now named exports so a sibling module can call them. `peg-policy-constants.mjs` exports eight previously private constants, `peg-policy-bucket.mjs` exports seven previously private functions, and `peg-policy-publication.mjs` exports four. No assertion, threshold, address, role, or error string changed. The dependency graph is acyclic and one-directional: constants → bucket → publication → runtime, with `runtime` holding `validatePegPolicyFoundation`, the single entry `index.mjs` calls.

### File map

| File | Lines | Change |
| --- | --- | --- |
| `scripts/lib/hcl.mjs` | 599 | moved, content identical |
| `scripts/lib/workflow-yaml.mjs` | 74 | new; 69 moved lines + import header |
| `scripts/production-infra-identity-contract/workflow-inventory.mjs` | 775 | was 844; lost the moved block, gained one import |
| `scripts/production-infra-identity-contract/peg-policy-constants.mjs` | 102 | new |
| `scripts/production-infra-identity-contract/peg-policy-bucket.mjs` | 460 | new |
| `scripts/production-infra-identity-contract/peg-policy-publication.mjs` | 442 | new |
| `scripts/production-infra-identity-contract/peg-policy-runtime.mjs` | 427 | new |
| `scripts/production-infra-identity-contract/peg-policy.mjs` | — | deleted (1,348 lines) |

Fifteen import-specifier edits across thirteen files. Five sit outside the contract directory: `check-metrics-bridge-template-plan.mjs`, `check-peg-policy-publication.mjs`, `deploy-staging-callsite-discovery.mjs`, `deploy-staging-contract.mjs`, and `sentry-provider-contract.test.mjs`. The shorter path let Prettier collapse one import in `check-metrics-bridge-template-plan.mjs` onto a single line; that is the only reformatting in the diff.

### Pins that must not move

`sanitize-terraform-output.sh` and `verify-github-environment-protection.mjs` stay flat, and the seven contract files that pin their exact workflow run-strings keep those pins verbatim. Proof: `git grep -c` for both names over the tree is byte-identical before and after this change (23 files, same counts).

### Routing

Three enumerated terraform paths-filters — `ci.yml` `terraform`, `infra.yml` push, `infra.yml` `pull_request` — cover the contract with the recursive `scripts/production-infra-identity-contract/**`. That prefix does not reach `scripts/lib/`, so each filter now also names `scripts/lib/hcl.mjs` and `scripts/lib/workflow-yaml.mjs`. Without this, a change to the shared HCL tokenizer would stop firing the terraform jobs it decides.

`agent-quality-gate.sh` gains `scripts/lib/*.mjs` on its contract-surface arm. `routing.test.mjs` now asserts all four surfaces, and the assertions are live — see Validation.

### Spec corrections

The phase spec was wrong on four points; the code follows the source, not the spec.

- **Five external `hcl.mjs` importers, not four.** `scripts/sentry-provider-contract.test.mjs` also imports it. It is a Sentry suite, so the manifest and suite gate cover it; both stay green.
- **`commonApplySteps` is not generic and did not move.** The spec put the workflow-yaml seam at L1-410, which would have carried `commonApplySteps` into `lib/`. That function hardcodes the `production-infra` environment name, the GCP WIF variables, pinned action SHAs, and the `verify-github-environment-protection.mjs` run-string. Moving it would relocate a security pin out of the contract. The real seam is L195-263: the six helpers with no policy in them.
- **`workflow-inventory.mjs` L412-675 did not fold into the peg-policy split.** That range ends in `protectedApplyJobInventory`, which serves all five protected apply workflows and is called by `workflow-inventory.mjs` itself, plus `fixtures.mjs`, `workflow.mjs`, and `refresh-routing.mjs`. Folding it would invert a dependency inside a fail-closed contract and file general apply machinery under a peg-policy name. It stays put.
- **The gate arm is not the only thing routing `pnpm tf:test`.** An unconditional sweep near the end of `agent-quality-gate.sh` runs the suite for every non-empty change set in the real tree, so the arm sets the reason string rather than deciding whether the suite runs. The negative controls below test the reason, which is what actually pins the arm.

Two further notes for the reviewer. The `peg-policy-publication.yml` `pull_request` paths list names `check-peg-policy-publication.mjs` but has never named the HCL parser that script reads through — that gap predates this PR and widening it means editing a pinned workflow inventory, so it is left alone. And `scripts/` sits outside `file-size-watchlist.mjs`'s scan scope, so the 600-line soft cap is a convention here, not a gate; all four new modules are well under it anyway.

### Architecture decision

No new ADR. This implements the structure ADR 0064 already decided, and touches ADR 0053 and ADR 0061 consumers without changing their semantics: the moved parsers carry no policy, and every assertion in those contracts is unchanged. ADR 0053, ADR 0055, and ADR 0061 gain evidence lines pointing at the new paths.

## Validation

All commands run on the branch head unless marked otherwise.

| Check | Result |
| --- | --- |
| `pnpm tf:test` | 10 suites green, same list as on `origin/main` before the change |
| `node .../index.test.mjs` | green (6 aggregated suites) |
| `node scripts/deploy-staging-contract.test.mjs` | `deployment source-staging contract tests passed` |
| `node .../security.test.mjs` via `index.test.mjs` | `production infrastructure identity security tests passed` |
| `node scripts/sentry-provider-contract.test.mjs` | `Sentry provider contract tests passed` |
| `node scripts/check-peg-policy-publication.test.mjs` | `peg policy publication boundary tests passed` |
| `node scripts/check-metrics-bridge-template-plan.test.mjs` | `Metrics Bridge template plan tests passed.` |
| `node scripts/check-deploy-root-anchors.test.mjs` | `All 4 deploy scripts anchor repo commands after deploy-guard.` |
| `node scripts/check-sentry-suites-in-ci.test.mjs` | `fail 0` |
| `node scripts/sentry-suite-gate.mjs` | 14 suites ran and were asserted |
| **Pin counts** | `git grep -c` output for `sanitize-terraform-output.sh` and `verify-github-environment-protection` is `diff`-identical before and after |
| **Verbatim proof** | every moved region matches its `origin/main` slice line for line |
| `pnpm agent:quality-gate --run` | green |
| `pnpm agent:quality-gate:test` | `agent quality gate tests passed` |
| `pnpm lint:scripts` | clean |
| `pnpm docs:index` | no diff |
| `pnpm agent:context-budget --strict` | `[OK] scripts — 16,462 bytes (86.9%), headroom 2,482` |
| `./tools/trunk check --all` | `Checked 2169 files ✔ No issues` |

### Negative controls

Each one was run, observed to fail, then reverted.

1. **Gate arm.** With `scripts/lib/*.mjs` removed from the contract-surface arm, a dry-run over `scripts/lib/hcl.mjs` reports `pnpm tf:test (non-empty change set validates production infrastructure contract)`. With the arm present it reports `pnpm tf:test (production infrastructure identity contract surface changed)`. The arm is what produces the difference.
2. **Gate arm, through the test.** The same removal fails `routing.test.mjs` with `scripts/lib/hcl.mjs must route pnpm tf:test as a contract surface, not through the unconditional sweep`.
3. **CI filter.** Deleting `- scripts/lib/hcl.mjs` from the `ci.yml` terraform filter fails `routing.test.mjs` with `ci.yml terraform filter must include scripts/lib/hcl.mjs`.
4. **Infra filter.** Deleting `- scripts/lib/workflow-yaml.mjs` from the `infra.yml` push block fails `routing.test.mjs` with `infra.yml push paths must include scripts/lib/workflow-yaml.mjs`. This also proves the test reads the push and `pull_request` blocks separately rather than matching one twice.

## Deferrals

- None

## Checklist

- [x] Architecture decision? Not applicable — this implements ADR 0064's recorded decision and changes no semantics in ADR 0053, ADR 0055, or ADR 0061.

---

> [!NOTE]
> **Medium Risk**
> Large refactor across fail-closed production identity and peg-policy contracts; behavior is intended to be byte-identical, but routing/CI wiring mistakes could let parser changes skip required checks.
> 
> **Overview**
> Implements ADR 0064 by moving **policy-free parsing** into `scripts/lib/`: `hcl.mjs` (Terraform HCL) and new `workflow-yaml.mjs` (Actions workflow / shell-run helpers). Callers in the production-infra identity contract, deploy-staging contract, and related checks now import `../lib/` instead of living under `production-infra-identity-contract/`.
> 
> The monolithic **`peg-policy.mjs`** is removed and split into **`peg-policy-constants`**, **`peg-policy-bucket`**, **`peg-policy-publication`**, and **`peg-policy-runtime`**, with **`index.mjs`** still exposing **`validatePegPolicyFoundation`**. Assertions and IAM rules are described as unchanged; the split mainly widens exports between sibling modules.
> 
> **CI and local gates** are updated so edits to the shared cores still run terraform/infra jobs and **`pnpm tf:test`**: `ci.yml` and `infra.yml` path filters list `scripts/lib/hcl.mjs` and `scripts/lib/workflow-yaml.mjs`; **`agent-quality-gate.sh`** routes **`scripts/lib/*.mjs`** as a contract surface; **`routing.test.mjs`** asserts those filters and routing reasons. ADRs 0053, 0055, and 0061 and **`scripts/AGENTS.md`** document the new layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d55e039175f2ce0cd0240d5085f2c884fc115790. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chapati23:
The auto-review failure above is the known CI reviewer outage (token resets ~2026-08-20), tracked in the reorg programme context. Both local review engines ran clean on this diff before push: codex 0.91, claude 0.75, dispositions in the PR body. No reviewer action needed on that failure.

PR #1928, opened 2026-08-18
feat(scripts): widen deploy routing globs for scripts/deploy/ (D1)

chapati23:
## The Problem

- Three routing arms in two trust roots match deploy scripts by a glob anchored on a literal prefix at the top of `scripts/`. All three stop matching when a wrapper moves into `scripts/deploy/`, which the scripts reorg (issue 1877) will do.
- Nothing fails when they stop. The root-anchor check and the Terraform/Cloud Run checklist just stop being scheduled, and the run still reads as green.

## The Solution

- Add the paired one-level arm ADR 0064 prescribes next to each pattern, keeping the flat pattern matching. A wrapper under `scripts/deploy/` now routes exactly what its flat counterpart routes today.
- Pin every case in both suites, each with a negative control, so a later edit that narrows the routing reds instead of going quiet.

## Details

Groundwork only. No file moves; the move is a separate PR.

Three arms changed, all additive:

| File | Arm | Now |
| --- | --- | --- |
| `agent-quality-gate.sh` | `scripts/deploy-*.sh` | `scripts/deploy-*.sh\|scripts/*/deploy-*.sh` |
| `agent-quality-gate.sh` | `scripts/deploy-indexer-logs.sh` | `…\|scripts/*/deploy-indexer-logs.sh` |
| `agent-quality-gate.sh` | `scripts/deploy-bridge.sh` | `…\|scripts/*/deploy-bridge.sh` |
| `agent-autoreview.sh` | `scripts/deploy-*.sh` in `select_checklists()` | `…\|scripts/*/deploy-*.sh` |

**Why the two exact-path arms are in scope.** Widening only the glob makes the
`deploy-indexer-logs.sh` case worse, not better: its exact-path arm sits first
in the same `case`, so after a move it stops matching, the widened glob catches
the path instead, and the run keeps scheduling the root-anchor check while the
runtime-log filter check drops out. Partial routing reads as working routing.
`deploy-bridge.sh` sits in a second `case` statement and so is not shadowed, but
it loses its Cloud Run checklist and `agent:context-check` on the same move.
Pairing both is what makes "a moved deploy script routes identically" true.

**Why both widened arms stay `.sh`-scoped.** The subject set of
`check-deploy-root-anchors.test.mjs` is `deploy-*.sh` files that source
`lib/deploy-guard.sh`; a Node helper has nothing for it to assert. The Node
deploy helpers landing in the same directory (`deploy-indexer-perf.mjs`,
`deploy-indexer-verify.mjs`, `filter-envio-runtime-errors.mjs`) drive Envio and
own no Cloud Run or Terraform surface, and are not routed the Cloud Run
checklist today. Both suites carry a Node-helper negative control pinning this.

**One live path newly reached.** `*` matches `/` in a `case` pattern, so the
pair reaches a `deploy-*.sh` basename under any `scripts/` subdirectory, today
`scripts/lib/deploy-guard.sh`. That breadth is deliberate:
`collectDeployWrappers()` in the check walks `scripts/` recursively, so routing
pinned to one fixed directory would again be narrower than the check it
schedules. The guard is the file every wrapper sources before it mutates
anything, which is exactly when both the check and the Cloud Run checklist
should run. Measured, the reach is strictly additive — the guard keeps
everything it routed before — and both suites pin that.

`scripts/AGENTS.md` needed no change and is untouched at 9,161 bytes, inside the
9,216-byte cap. ADR 0064 is updated: it named these globs as the unfixed case,
and its sweep checklist told a later move to add the pair that now exists.

## Validation

**Quality gate** — `pnpm agent:quality-gate --run --parallel 4 --skip-if-fresh --base origin/main`, all 12 mapped commands passed:
`agent:quality-gate:test` (7m53s), `agent:autoreview:test` (4m35s, all five families),
`docs:index --check`, `agent:context-check`, `check-sentry-suites-in-ci.test.mjs`,
`docs:navigation-eval --check-fixtures`, `tf:test`, Trunk on all five files, and `bash -n` on all four shell files.

**Negative controls, both proven by reverting the widening and restoring it:**

- Gate: with `scripts/*/deploy-*.sh` removed, `agent-quality-gate.test.sh` fails —
  `expected output to contain: - node scripts/check-deploy-root-anchors.test.mjs (deploy wrapper changed)`,
  with the gate output showing `scripts/deploy/deploy-indexer.sh` mapping only Trunk and `tf:test`. Restored: suite passes.
- Autoreview: with the alternative removed, `AUTOREVIEW_TEST_FOCUS=bundle-integrity` fails —
  the nested bundle's `selected-checklists.txt` holds only `recurring-review-patterns.md`,
  `review-prompt-exclusions.md`, and `code-health.md`; `terraform-cloudrun.md` is absent. Restored: family passes.

**Shadowing check** (the review question that drove the two exact-path pairings) — routing for
`scripts/lib/deploy-guard.sh` compared with the widened alternatives present and removed: identical apart
from the added root-anchor check, so nothing was displaced. `scripts/deploy/deploy-indexer-logs.sh` keeps
both its commands, and `scripts/deploy/deploy-bridge.sh` keeps all three.

**Local autoreview, both engines clean** (review-bot substitute while both bots are down). Because this PR
changes the autoreview runtime itself, the adapter fail-closes on self-review; both runs used the runbook's
trusted-checkout procedure — a clean `origin/main` clone with an explicit `AUTOREVIEW_HELPER`, verified
byte-identical to protected main for all three runtime files.

- codex: `autoreview clean: no accepted/actionable findings reported`, `patch is correct (0.94)`
- claude: `autoreview clean: no accepted/actionable findings reported`, `patch is correct (0.75)`

Earlier rounds were not clean, and both fixes are in the diff: the claude engine flagged the undocumented
reach to `scripts/lib/deploy-guard.sh` (pinned in both suites, documented in the ADR) and then the
exact-path shadowing hazard (both arms paired, full command sets asserted). Its remaining reservations —
that the "new `scripts/<dir>/deploy-*.sh` arm goes above the pair" invariant is comment-enforced, and that
dropping the `deploy-` basename prefix would defeat both arms — it filed as non-blocking; ADR 0064 already
mandates keeping the prefix, and both belong to the move commit.

## Checklist

- [x] Architecture decision? No new decision; ADR 0064 is updated to match the code it governs.

---

> [!NOTE]
> **Medium Risk**
> Changes CI/review routing for deploy and Cloud Run paths; behavior is additive and heavily regression-tested, but mis-routing would fail open (skipped checks) rather than fail closed.
> 
> **Overview**
> Prepares the upcoming `scripts/deploy/` reorg by widening **deploy script routing** so checks do not silently drop when wrappers move one directory down.
> 
> **Quality gate** (`agent-quality-gate.sh`) pairs `scripts/*/deploy-*.sh` (and nested exact paths) with existing flat patterns for the deploy glob, `deploy-indexer-logs.sh`, and `deploy-bridge.sh`, preserving the full command set—including cases where a widened glob alone would shadow a more specific arm and drop checks.
> 
> **Autoreview** (`agent-autoreview.sh`) applies the same paired pattern so the Terraform/Cloud Run checklist still selects for nested `deploy-*.sh` paths.
> 
> **Tests** add nested-path coverage in `agent-quality-gate.test.sh` and `run_deploy_directory_checklist_routing_regression` in `agent-autoreview.test.sh`, including negative controls (Node helpers stay shell-scoped; `scripts/lib/deploy-guard.sh` intentionally gets deploy checks).
> 
> **ADR 0064** documents that deploy pairs land before the move and updates the sweep checklist accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba3de56db7812cd01c1b3904825d9f05118414bf. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment script detection for nested paths.
  * Ensured deployment-related checks are correctly applied to shell wrappers while excluding unrelated helpers.

* **Tests**
  * Added regression coverage for nested deployment routing.
  * Verified Cloud Run, context, runtime-log, root-anchor, and shell syntax checks remain correctly assigned.

* **Documentation**
  * Updated architecture documentation with deployment routing rules, checklist requirements, and validation evidence.

### DIFF
--- a/docs/adr/0064-scripts-module-directories.md
+++ b/docs/adr/0064-scripts-module-directories.md
@@ -3,7 +3,7 @@ title: scripts/ may use module subdirectories; basenames and pinned paths are th
 status: active
 owner: eng
 canonical: true
-last_verified: 2026-08-17
+last_verified: 2026-08-18
 scope: ci/process
 date: 2026-08
 doc_type: adr
@@ -62,6 +62,27 @@ already carries the paired arm `scripts/*/sentry-*.test.mjs` next to
 all 53 `sentry-*` files wherever they land, keep the `deploy-` prefix on the
 deploy wrappers, and add the paired one-level arm whenever a literal-prefix glob
 is the routing.
+
+Both deploy arms now carry that pair, added ahead of the move rather than with
+it: `scripts/deploy-*.sh|scripts/*/deploy-*.sh` routes the root-anchor check in
+`agent-quality-gate.sh`, and the same pair routes the Terraform/Cloud Run
+checklist in `select_checklists()` in `agent-autoreview.sh`. Both stay
+shell-scoped. The check behind the first has a `deploy-*.sh` subject set, and
+the Node deploy helpers that land in the same directory drive Envio and own no
+Cloud Run surface, so neither arm is the right home for them. Widening a routing
+glob is separable from the move it protects, and doing it first means the move
+cannot be the commit that goes quiet.
+
+Because `*` matches `/`, the pair reaches a `deploy-*.sh` basename under any
+`scripts/` subdirectory rather than one fixed directory. That is deliberate on
+both arms: `collectDeployWrappers()` in `check-deploy-root-anchors.test.mjs`
+walks `scripts/` recursively, so routing pinned to a single directory would once
+again be narrower than the check it schedules. The one live path it newly
+reaches is `scripts/lib/deploy-guard.sh`, the guard every wrapper sources before
+it mutates anything, which is precisely when both the check and the Cloud Run
+checklist should run. Both suites pin it. One consequence for later edits: a
+`case` takes the first matching arm, so a new arm for a path of the shape
+`scripts/<dir>/deploy-*.sh` goes ABOVE the pair or it never runs.
 
 `scripts/sentry-suite-manifest.json` is stricter than a glob. Its keys are exact
 repo-relative paths, and `scripts/sentry-suite-gate.mjs` reconciles them against
@@ -179,7 +200,16 @@ routing, not procedure.
 8. `docs/notes/quick-commands.md`.
 9. `agent-quality-gate.sh` routing arms — a literal-prefix glob such as
    `scripts/deploy-*.sh` or `scripts/sentry-*.test.mjs` stops matching one
-   directory down. Keep the basename prefix; add the paired one-level arm. Its
+   directory down. Keep the basename prefix; add the paired one-level arm. The
+   `sentry-` and `deploy-` arms already carry theirs, so a move of those files
+   verifies the pair rather than adding it. An arm naming an exact path is a
+   literal, not a glob, and needs the same pairing for the same reason: all
+   three on the deploy leg — `scripts/deploy-indexer-logs.sh`,
+   `scripts/deploy-bridge.sh`, and the `deploy-*.sh` glob — now carry it, so a
+   moved wrapper keeps its whole command set. Leaving one unpaired below a
+   widened glob is the worst case, not the safe one: the glob catches the moved
+   path, the run still looks routed, and only the arm's extra commands go
+   missing. Its
    contract-surface arm also names `scripts/lib/*.mjs`, which sets the
    `pnpm tf:test` reason; the unconditional sweep already runs the suite. Its
    `implementation_signature()` path list is stricter than a glob: an entry it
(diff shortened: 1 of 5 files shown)


chapati23:
Replies to the two CodeRabbit nitpicks (cr-comment ids 7b649997d008a8d79b79f471, aba067bd9ed93fb49feb1e77):

**Reduce duplicated routing rationale — won't fix.** Each inline block states a constraint an arm editor needs at edit time: the first-match ordering invariant (a new `scripts/<dir>/deploy-*.sh` arm goes above the pair or never runs), the shell-only scoping and why, and the any-depth reach with its one live consequence. ADR 0064 holds the canonical narrative; the arm comments are the operational subset for editors who do not open the ADR, which is the same failure mode this PR closes at the routing layer. This convention is consistent across the reorg programme's trust-root arms (see the paired sentry arms and their comments).

**Two-level nested regression — deferred to D2 with a named owner, not dropped.** The claim "reaches every depth" is currently pinned at one level. D2 (the scripts/deploy/ move, tracked on issue 1877) already adds nested routing assertions for the exact-path .mjs arms; a two-level probe case (`scripts/deploy/region/deploy-probe.sh` asserting both the generic and a specialized command, plus the autoreview checklist twin) is added to that PR's assertion list, where it rides an already-required validation round instead of re-running this PR's full gate for a fixture-only change. No real path at two levels exists or is planned; the case pins the `case`-pattern mechanism only.

Issue #1929, opened 2026-08-18
sentry triage pipeline: three pre-existing fail-open gaps found during D4 directory move review


chapati23:
## Context

While reviewing the D4 sentry directory move (issue 1877, `scripts/sentry-*` files
into `scripts/sentry/{triage,autofix,gate,broker,ci-wiring}/`), codex and claude
autoreview passes on the split commits surfaced three findings in code the move
did not modify. Each is confirmed pre-existing: `git diff` of the file at its old
flat path (origin/main) against its new nested path shows only doc-comment path
text changed, with the flagged logic byte-identical.

All three sit in the triage/re-queue verification path and share a theme: they
weaken a fail-closed guarantee to fail-open under specific conditions. None
blocks the D4 move; filing here per the deferred-work convention.

## 1. `sentry-triage-ingest.mjs` — pagination loop truncates silently past `maxPages`

`scripts/sentry/triage/sentry-triage-ingest.mjs` (~line 536)

The Sentry API pagination loop stops after `maxPages` even when the final
response still reports `next.hasResults`. The function returns the partial
result set without raising an error, unlike `ghPaginate`, which does throw in
the equivalent situation.

Codex (P2, confidence 0.98): "throw when a next-page URL still exists because
the page limit was reached."

## 2. `sentry-triage-requeue.mjs` — `isSelectableForTriage` accepts any non-CLOSED state

`scripts/sentry/triage/sentry-triage-requeue.mjs` (~line 241)

```js
export function isSelectableForTriage({ state, labels } = {}) {
  const names = Array.isArray(labels) ? labels : [];
  return (
    String(state ?? "").toUpperCase() !== "CLOSED" &&
    names.includes(QUEUE_LABEL) &&
    names.includes(NEEDS_TRIAGE_LABEL)
  );
}
```

If GitHub returns an incomplete or unexpected payload, `readStubState`
normalizes a missing state to `""`, which is not `"CLOSED"` — so the end-state
verifier can report success without having confirmed the issue is actually
OPEN. Codex (P2, confidence 0.96): require
`String(state).toUpperCase() === "OPEN"` explicitly instead of `!== "CLOSED"`.

## 3. `sentry-triage-workflow-requeue.mjs` — terminal check races archive settlement

`scripts/sentry/triage/sentry-triage-workflow-requeue.mjs` (~line 246)

The terminal-state check the re-queue compensation runs before reopening a stub
[...]

PR #1930, opened 2026-08-18
refactor(scripts): move sentry tooling into scripts/sentry/ (D4)

chapati23:
## The Problem

- 65 sentry-prefixed files sit flat at the top of `scripts/`, spanning five
  distinct subsystems (CI-wiring checks, the self-run suite gate, the autofix
  pipeline, the triage pipeline, and the MCP credential broker).
- A flat layout makes ownership and blast radius illegible at a glance and
  blocks the directory-scoped conventions the rest of `scripts/` is moving to
  (issue 1877, ADR 0064).

## The Solution

- Move all 65 files plus `scripts/static-imports.mjs` (which the gate's
  dependency-freedom proof couples to) into `scripts/sentry/{ci-wiring,gate,
  autofix,broker,triage}/` and `scripts/lib/`, preserving every basename.
  Repoint every importer, workflow, routing arm, and doc that named the old
  flat paths.

## Details

**Target layout** (65 files + `scripts/static-imports.mjs`):

| Directory                  | Files | Contents                                              |
| --------------------------- | ----- | ------------------------------------------------------ |
| `scripts/sentry/ci-wiring/` | 12    | `check-sentry-suites-in-ci*`                            |
| `scripts/sentry/gate/`      | 9     | `sentry-suite-gate*`, the manifest, the provider-contract exemption fixture |
| `scripts/sentry/autofix/`   | 18    | `sentry-autofix-*`                                      |
| `scripts/sentry/triage/`    | 24    | `sentry-triage-*`                                       |
| `scripts/sentry/broker/`    | 2     | `sentry-mcp-broker.mjs` + test                          |
| `scripts/lib/`              | +1    | `static-imports.mjs` (moved with the gate; the gate's dependency-freedom check imports it) |

**Plan discrepancies found, and how they were resolved:**

- The plan estimated 19 autofix files; the actual count is 18. No file is
  missing — the plan's number was off.
- One file, `sentry-provider-contract.test.mjs`, didn't belong to any of the
  five named families. It's the suite the gate exempts and routes to
  `production-infra-contract` instead of running directly (it's
  `EXEMPT_SUITE` in `sentry-suite-gate-manifest.mjs`), so it went to
  `scripts/sentry/gate/` alongside the mechanism that exempts it.
- Every file that computes its own location from `import.meta.url` (or a
  `path.resolve(dirname, "..")`-style ancestor walk) needed its ancestor
  count bumped by two, since files now sit two directories deeper. This
  category wasn't called out explicitly in the move plan; it touched
  `sentry-suite-gate.mjs`, its fixtures and integrity/isolation tests, the
  CI-wiring probes, the broker test, and several triage tests. All were found
  and fixed; one (`sentry-triage-requeue.test.mjs`'s "exactly one call site"
  scan) shipped with one `..` too many during development and was caught by
  autoreview before push — see Validation.
- `verifyExemptRoute` in `sentry-suite-gate.mjs` assumed the exempt suite and
  its importer always shared a directory (building the expected import
  specifier from a bare basename). That's no longer true once the suite
  moved under `scripts/sentry/gate/` while its importer
  (`scripts/tf-stacks.test.mjs`) stayed flat. Generalized it to compute the
  specifier via `relative(dirname(importer), suite)`.
- Two non-recursive `readdirSync` directory scans (in triage tests proving
  "this function has exactly one call site across `scripts/`") assumed a flat
  `scripts/` layout and would have silently scanned a fraction of the tree
  post-move. Both were rewritten to walk `scripts/` recursively
  (`{recursive: true, withFileTypes: true}`), with an anti-vacuity floor
  (`scanned >= N`) so a future regression to a narrow scan fails loudly
  instead of passing having checked almost nothing.
- `scripts/AGENTS.md`'s "phases P1-P13" note became "P1-P14" — inferred as
  the next sequential phase number; no explicit "P14" label existed anywhere
  in the repo to confirm against, so flagging this as an inference rather
  than a verified fact.

**Invariants honored:**

- Every moved file's basename is byte-identical to its pre-move name; only
  the directory changed (`git diff -M50%` shows 65 clean renames).
- The manifest (`scripts/sentry/gate/sentry-suite-manifest.json`) still
  matches the on-disk suite set by exact equality in both directions; the
  negative control (deleting one manifest entry) reproduces the intended
  ROUTE-BROKEN/set-mismatch failure and is restored.
- The `sentry-suites` CI job stays the 5th required, unconditional status
  context — byte-identical name, no `if:`, still in `ci`'s `needs` and absent
  from `allowed-skips`. All `.github/workflows` references to the moved
  paths were swept and repointed.
- `findSentrySuites()` and the CI-wiring checker's enumeration already walked
  `scripts/` recursively before this PR (confirmed via inspection), so
  discovery needed no change; only fixed-depth ancestor-path arithmetic
  needed correction (see discrepancies above).
- Cross-family imports were repointed; same-family (same-directory) imports
  needed no change. `package.json`'s 16 `sentry:*` scripts and their mirror
  in `scripts/check-agent-quality-gate-package-scripts.mjs` (the package-script
  alias registry) were repointed in lockstep and verified to match key for
  key.
- `agent-quality-gate.sh`'s per-file sentry routing arms were repointed; the
  two generic wildcard arms (`scripts/sentry-*.test.mjs` and
  `scripts/*/sentry-*.test.mjs`) needed no change — in a bash `case`, `*`
  matches `/`, so both arms already reach any depth.
- `scripts/AGENTS.md` stays under its 9216-byte budget (measured 9198 with
  `wc -c` after this PR's edits).
- Docs were swept per ADR 0064's module-directory checklist: `scripts/AGENTS.md`,
  ADR 0038, 0056, 0062, 0064, 0065, `docs/notes/sentry-triage-pipeline.md`,
  `docs/notes/agent-quality-gate-mechanics.md`, `docs/notes/file-size-watch.md`,
  `alerts/AGENTS.md`, `alerts/infra/README.md`.

**Review-driven fix (introduced by this move, caught before push):**

- `sentry-triage-requeue.test.mjs`'s recursive scan for
  `buildRegressedComment`'s single call site computed its root with three
  `..` segments from the file's new location (landing at the repo root
  instead of `scripts/`). The floor assertion (`scanned >= 60`) still passed
  because scanning the whole repo trivially exceeds it, masking the bug —
  it silently walked `node_modules` and every other package instead of just
  `scripts/`. Caught by claude-engine autoreview; fixed to two `..` segments,
  matching the sibling `sentry-triage-brief.test.mjs`'s correct pattern.

## Validation

- `git diff -M50% --name-status origin/main`: 65 R (clean renames) + 1 A/D
  pair (the manifest; its content changed enough that git's default
  similarity heuristic didn't call it a rename) + 23 M (pin repoints).
- Repo-wide grep for stale flat `scripts/sentry-*`/`scripts/check-sentry-suites-in-ci*`
  references (excluding the intentional generic-wildcard prose in ADR 0064
  explaining the routing-arm mechanism, and one deliberate
  `scripts/sentry-not-yet-written.test.mjs` fixture in
  `agent-quality-gate.test.sh` that tests the flat wildcard arm's own
  routing): zero remaining.
- Manifest negative control: removing one entry from
  `scripts/sentry/gate/sentry-suite-manifest.json` reproduces the intended
  set-mismatch failure; restoring it returns the gate to green.
- `node scripts/sentry/gate/sentry-suite-gate.mjs`: all 14 non-exempt suites
  pass from their own snapshot's output (pass counts match or exceed their
  manifest floors); the 1 exempt suite's route to `production-infra-contract`
  is verified without running it.
- `node scripts/sentry/ci-wiring/check-sentry-suites-in-ci.test.mjs`: 43/43
  passed, including the suite-enumeration-walks-symlinks and
  file-size-hard-cap checks.
- `pnpm docs:index --check`: clean.
- `pnpm agent:quality-gate --run --parallel 4 --skip-if-fresh --base origin/main`
  (and the pre-push hook's own `--parallel 3` invocation): all mapped
  commands passed, including every sentry suite, `tf-stacks.test.mjs`
  (exempt-route consumer), `agent-quality-gate.test.sh` (the gate's own
  self-test suite, ~8 min), and the full turbo lint/typecheck/knip sweep
  across the affected packages.
- Both local autoreview engines (bots are down until ~2026-08-20 — issue
  1917/ADR 0066 CodeRabbit cutover) ran clean across all 11 per-commit
  reviews that carry code changes:
  - codex: every commit reviewed individually. Two categories of finding
    surfaced, both resolved: (a) references to sibling files not yet
    present at an intermediate commit — an expected artifact of splitting
    the move into add-only, individually-reviewable commits (each commit's
    bundle otherwise exceeded the autoreview tool's 512,000-byte
    single-pass limit); the branch HEAD is complete and every suite passes
    from HEAD. (b) three genuine pre-existing bugs, confirmed via `git diff`
    against `origin/main` to be byte-identical apart from doc-comment path
    text — filed as issue 1929, not fixed here (out of scope for a pure
    move). One genuine review-caught regression (the `requeue.test.mjs`
    scan-root bug above) was fixed before push.
  - claude: same pattern — split-commit-context findings (self-resolving
    at HEAD, individually verified via direct test execution and `git diff`
    against `origin/main`) plus the one genuine finding already covered
    above.
  - Full-branch-mode autoreview (both engines) fails outright on this diff
    with a tool capacity limit unrelated to content — filed as issue 1931.
    See Deferrals.
- Sentry-suite-gate test count: 14 non-exempt suites ran before this PR
  (measured against `origin/main`'s manifest) and 14 run after (this PR only
  moves files; it adds no suite and removes none).
- CodeRabbit's automated review on this PR flagged several findings, all in
  files this PR only relocates (verified byte-identical to `origin/main`
  apart from doc-comment path text): `spawnSync` `killSignal` in
  `check-sentry-suites-in-ci-gate-extract.mjs`, two broker robustness
  nitpicks in `sentry-mcp-broker.mjs`, an unused `now` option in
  `sentry-triage-digest.mjs`, and a dropped test-helper message argument in
  `sentry-autofix-run-record.test.mjs`. Fixing pre-existing behavior here
  would break this PR's verbatim-move proof discipline, so all five are
  appended to issue 1929 for a focused follow-up PR instead.

## Deferrals

- https://github.com/mento-protocol/monitoring-monorepo/issues/1929 — three pre-existing, unrelated-to-this-move bugs found via autoreview in the sentry-triage state-verification path (pagination truncation, a state-check gap, a cross-workflow TOCTOU race), plus five further pre-existing findings from CodeRabbit's review of this PR (gate-extract killSignal, two broker nitpicks, digest unused option, assertEqual dropped message). Fixing pre-existing behavior here would break this PR's verbatim-move proof discipline, so all are routed to a focused follow-up PR instead.
- https://github.com/mento-protocol/monitoring-monorepo/issues/1931 — the autoreview tool's branch-mode diffing hardcodes `--no-renames` at every call site, so a rename-heavy branch like this one blows the 4,096,000-byte aggregate review-input cap before either engine's analysis even starts. Worked around here with a per-commit review split (see Validation). This PR holds for that fix to land on `main`, then re-merges `main` before ship.

## Checklist

- [x] Architecture decision? This PR executes the layout ADR 0064 already
      decided and endorsed (including this exact one-directory-per-family
      convention under `scripts/sentry/`); it makes no new architectural
      decision, so no new ADR is added.

## Summary by CodeRabbit

* **New Features**
  * Added automated Sentry triage digesting and issue projection, including duplicate prevention, validation, label handling, and Slack-ready summaries.
  * Added a restricted credential broker for safer Sentry integrations.
  * Added stronger text sanitization and validation for generated issue content.

* **Bug Fixes**
  * Improved handling of malformed, duplicate, stale, or failed triage items.

* **Tests**
  * Expanded CI and quality checks, including configuration validation and workflow integrity coverage.

* **Documentation**
  * Updated Sentry automation runbooks and architecture references.


chapati23:
Thanks for the review. This PR intentionally changes zero behavior — it is a pure directory move with byte-identical file content (only doc-comment path references and two token-placeholder fixtures for a local-scanner false positive were edited; verified per-file against `origin/main` with `git diff`). Every content-level finding below lands on a file this PR only relocates, so it is by definition pre-existing, not introduced here:

- `check-sentry-suites-in-ci-gate-extract.mjs:56-70` — `spawnSync`'s `killSignal` should be `SIGKILL` so a timed-out probe that ignores `SIGTERM` still terminates. Valid finding.
- `sentry-mcp-broker.mjs:374-380` — the response relay drops `retry-after` and the Sentry rate-limit headers, so the MCP client can't back off on a 429. Valid finding.
- `sentry-mcp-broker.mjs:406-409` — the startup `error` listener stays attached after `listen` resolves, so a later runtime socket error is silently discarded instead of logged. Valid finding.
- `sentry-triage-digest.mjs` (~684-687) — an unused `now` option is passed into `buildDigest`. Valid finding.
- `sentry-autofix-run-record.test.mjs:26-32` — `assertEqual` accepts only two parameters, so the message argument at its one call site is silently discarded. Valid finding.

Fixing any of these here would break the verbatim-move proof this PR is built on (every moved file's content matches `origin/main` exactly, which is what the PR body's Validation section demonstrates). All five are appended to issue 1929, which already tracks other pre-existing findings surfaced during this move's review, for a focused follow-up PR.

Issue #1955, opened 2026-08-20
Close depletion-alert coverage gaps: EURm/EUROP 1:1 fallback, unread invertRateFeed visibility


chapati23:
## Problem

PR #1944 gives depletion alerting a value-weighted share gauge (`mento_pool_reserve_value_share_token{0,1}`), published only when the pool has a live median and a known `invertRateFeed`. Two coverage gaps remain, both deliberate fail-closed choices in that PR:

1. **EURm/EUROP (Polygon) has no depletion coverage.** The pair's oracle rate is manually hardcoded 1:1 (Chainlink has no EUROP/EUR feed), so the pool never lands a SortedOracles median and never gets a value share. At a true 1:1 rate, token-count share equals value share, so the existing count gauges are already the correct depletion measure for this one pool. Fix: per-pool fallback in metrics-bridge that publishes the value-share gauges from count shares when the rate feed is the hardcoded 1:1 case. Also verify the oracle-liveness rules treat this pool sensibly — "no median ever" is its permanent normal state, not an outage.

2. **A future pool with an unread `invertRateFeed` silently lacks depletion coverage.** The bridge fails closed (publishes nothing) when the flag is unavailable, which is correct but invisible. Fix: an alert on "FPMM pool has reserves but publishes no value share" (excluding the known 1:1 special case) so the gap surfaces instead of staying dark.

## Context

- ADR 0067 (depletion-based criticality) and PR #1944 document the fail-closed decisions.
- Origin: #1936 (critical alert threshold retuning).

chapati23:
Both gaps are closed and verified in production (2026-08-20):

1. **EURm/EUROP 1:1 fallback** — PR #1958 (merged, deployed via run 32382347241). The bridge now publishes `mento_pool_reserve_value_share_token{0,1}` for pool `137-0xcd8c…3201` from count shares, gated on allowlist membership + `lastMedianPrice == 0` (never priced) + `tokenDecimalsKnown`. Live check: value shares 0.40/0.60, and the pool appears in the depletion rules' input — 18/18 funded pools covered. Review also hardened the general path: `tokenDecimalsKnown` now gates `reserveValueShares` for every pool (unread decimals would misscale a 6-decimal leg by 10^12).

[...]

PR #1957, opened 2026-08-20
feat(alerts): surface FPMM pools missing value-share coverage

chapati23:
## The Problem

- The pool depletion alerts read the value-share gauges, which metrics-bridge publishes only for a pool with a live oracle median and an on-chain-read `invertRateFeed`. Correct, but silent.
- A pool that trips that gate leaves both depletion rules at NoData, and `no_data_state = "OK"` turns NoData into no notification. The pool has no depletion coverage and nothing says so.
- Today one pool is in that state (Polygon EURm/EUROP). A future pool with an unread `invertRateFeed` would join it invisibly.

## The Solution

- Add `Pool Value Share Missing` — a warning-tier rule that fires when a pool publishes token-count shares but no value shares. The count shares are the bridge's own proof the pool holds reserves, so the difference between the two gauge families is exactly "funded, but outside depletion coverage".
- It reports the gap; it does not fill it. Slack `#alerts-pools`, with a summary that points at the two things that cause it: an unread `invertRateFeed` or a rate feed with no live median.

## Details

**Expression** (`alerts/rules/main.tf`, `pool_value_share_missing_active_promql`):

```promql
max without(token_symbol) (mento_pool_reserve_share_token0 or mento_pool_reserve_share_token1)
  unless on(chain_id, pool_id, pair)
    (mento_pool_reserve_value_share_token0 or mento_pool_reserve_value_share_token1)
```

- Same unless-shape as `deviation_warning_unavailable_active_promql`, which covers the matching data-unavailable gap on the deviation side.
- `unless on(chain_id, pool_id, pair)`, not a bare `unless`: both gauge families carry `token_symbol` and their own `__name__`, so a full label-set comparison would never match and the rule would fire on every funded pool.
- `max`, not the `min` the issue sketched. On a fully one-sided pool the thin side reads `0.0`, and the rule's `> 0` threshold would then skip exactly the pool with the worst depletion exposure. The fat side is `>= 0.5` for any funded pool. No pool sits at a zero count share today, so this is defensive rather than a change to current behaviour — verified live, see Validation.
- `without(token_symbol)` drops `__name__` and `token_symbol`, leaving the pool fingerprint the alert instances, the Slack template, and the `notify_warning_pools_pool` `group_by` are keyed on.

**Rule placement** — third rule inside `grafana_rule_group.fpmms_depletion` rather than a sibling group. `fpmms_deviation` already sets the precedent: `Deviation Breach (anchored)` covers the window where `Deviation Breach`'s input is unavailable and lives in the same group as the rule it backs. A rule that reports another rule's coverage gap is read with that rule, and shares its 60s interval.

**Rule shape** — `severity = "warning"`, `no_data_state = "OK"` per fpmms convention, `exec_err_state = "Error"`, and the instant-query + `__expr__` threshold pair copied from the depletion rules with the `gt 0` condition shape of `Deviation Breach (anchored)`. `notification_settings` copies `local.notify_warning_pools_pool` verbatim.

- `for = "30m"`. The gap is structural — an unread orientation flag or a feed that has never landed a median stays that way until someone changes something — so nothing is lost by waiting, and the dwell absorbs a bridge restart or a run of missed scrapes that would otherwise read as a minute-long loss of coverage.
- No `keep_firing_for`: no warning rule in `rules-fpmms.tf` carries one.
- No annotation-only queries. The `V0`/`V1` queries the depletion bands attach select exactly the series this rule fires on the *absence* of, so attaching them would hand Grafana a guaranteed-empty query and risk NoData-ing the whole rule (`alerts/AGENTS.md`).
- `description` is written for the Grafana rule-detail view only — the Slack template renders `description` for critical severity alone, so everything an operator needs in the channel is in `summary`.

**Merge ordering.** Merge this after the sibling metrics-bridge PR (EURm/EUROP 1:1 fallback, also part of #1955) has **deployed**. Until then this rule fires on Polygon EURm/EUROP, which is a true positive under today's bridge — the pool genuinely has no value share — but it is the gap that PR closes. Landing this first costs one harmless `#alerts-pools` warning that self-resolves on that deploy. No exclusion for EURm/EUROP is encoded here on purpose: once the fallback ships, that pool publishes value shares and drops out of this expression by itself, and a hardcoded exclusion would have to be removed again.

**Not changed:** `check-deviation-threshold-drift.mjs`. This rule mirrors no numeric threshold into `shared-config` — the `> 0` is a presence test, not a tuned band — so there is nothing for the drift checker to compare. Confirmed by running it.

## Validation

Live check against production Mimir (`grafanacloud-prom`), which is the strongest evidence available for a PromQL change:

- `count(max without(token_symbol) (mento_pool_reserve_share_token0 or mento_pool_reserve_share_token1))` → **18** funded pools.
- The full rule expression → **exactly 1 series**: Polygon `EURm/EUROP`, value `0.600`. No false positives across the other 17 pools.
- Returned labels carry the full pool fingerprint (`chain_id`, `chain_name`, `pair`, `pool_id`, `pool_address_short`, `block_explorer_url`), so the Slack title link and the routing `group_by` both resolve.
- `count(min without(token_symbol) (... ) == 0)` → no series, confirming the `max`-over-`min` choice is defensive today rather than a behaviour change.

Local checks:

- `node scripts/alerts/alert-rules-lint.mjs` → `172 PromQL expressions parsed, 55 referenced metric names checked against 70 registered gauges, peg policy validated`. Both gauge families are already registered in metrics-bridge on `main`, so the metric cross-check passes without a bridge change.
- `node --test scripts/alerts/alert-rules-lint.test.mjs` → **50 passed**, including the new pin `a funded pool that publishes no value share surfaces as a warning` (both gauge families in the expression, the `unless on(chain_id, pool_id, pair)` join, the `max` aggregation, and severity / route / dwell / `no_data_state` / no-`keep_firing_for` / no-empty-annotation-query on the rule).
- `node scripts/alerts/check-deviation-threshold-drift.mjs` → `OK: tolerance=1.01, depletion critical=0.2, depletion page=0.1`. Unchanged, as expected.
- `terraform -chdir=alerts/rules init -backend=false && terraform validate` → **Success! The configuration is valid.**
- `terraform fmt -check` → clean. `./tools/trunk check` on the four changed files → no issues.
- `pnpm agent:quality-gate --run --parallel 4 --skip-if-fresh` → green.

**No plan diff in this PR.** PR-level Terraform plans are targeted and exclude the fpmms rule groups (they run against dummy secrets), so this change gets `terraform validate` plus the trusted-main plan only. The rule reaches Grafana through the alerts-rules workflow's apply-on-merge behind the production approval gate; there are no manual apply steps.

## Deferrals

- None

## Checklist

- [x] Architecture decision? Not applicable — this implements the visibility gap ADR 0067 already recorded; the ADR gets a short note that the silence is now surfaced.

Part of #1955

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 544806fb0e339c7e2dd91716c7a47066163087e3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added monitoring for funded pools with reserve share data but missing value share metrics.
  * Introduced a warning alert after 30 minutes when value-share coverage is absent.
  * Added operator guidance and resolution details for affected pools.
* **Documentation**
  * Documented the new alert and its routing.
* **Tests**
  * Added regression coverage for detection, severity, timing, and no-data behavior.

### DIFF
--- a/alerts/rules/main.tf
+++ b/alerts/rules/main.tf
@@ -325,6 +325,39 @@ locals {
   pool_depletion_page_active_promql     = local.pool_min_reserve_value_share_promql
   pool_depletion_critical_active_promql = "(${local.pool_min_reserve_value_share_promql}) >= 0.1"
 
+  # ── Value-share coverage gap ─────────────────────────────────────────────
+  # Both bands above read the VALUE-share gauges, which the bridge publishes
+  # only for a pool with a live median AND an on-chain-read `invertRateFeed`
+  # orientation. That gate is fail-closed and correct — a depletion number
+  # derived from a guessed orientation would be worse than none — but it is
+  # silent: a pool that trips it leaves both bands at NoData, and `no_data_state
+  # = "OK"` turns that into no notification at all. This expression is the
+  # visibility companion, the same unless-shape as
+  # `deviation_warning_unavailable_active_promql` above, which covers the
+  # matching data-unavailable gap on the deviation side.
+  #
+  # Left side — pools the bridge sees holding reserves. The count-share gauges
+  # publish whenever normalized reserves sum above zero, so their presence is
+  # exactly "this pool is funded" and needs no separate reserves metric.
+  #
+  # `max`, not `min`: on a fully one-sided pool the thin side reads 0.0, and a
+  # `> 0` threshold would then skip the pool with the WORST depletion exposure.
+  # The fat side is >= 0.5 for any funded pool, so `max` fires for every funded
+  # pool, and the value it fires with reads as the concentration on the heavy
+  # leg. The alert is a presence signal; the number is diagnostic only.
+  #
+  # `without(token_symbol)` folds the two flat per-token series into one series
+  # per pool and drops `__name__`, leaving exactly the pool fingerprint the
+  # alert instances and the `notify_warning_pools_pool` `group_by` are keyed on
+  # — same reason `pool_min_reserve_value_share_promql` aggregates that way.
+  #
+  # Right side — `unless on(chain_id, pool_id, pair)`, not a bare `unless`: the
+  # value gauges carry `token_symbol` and their own `__name__`, so a full
+  # label-set comparison would never match and the rule would fire on every
+  # funded pool. `on()` restricts the join to the pool fingerprint the two
+  # gauge families share.
+  pool_value_share_missing_active_promql = "max without(token_symbol) (mento_pool_reserve_share_token0 or mento_pool_reserve_share_token1) unless on(chain_id, pool_id, pair) (mento_pool_reserve_value_share_token0 or mento_pool_reserve_value_share_token1)"
+
   # Transition markers let resolved notifications say why an alert stopped
   # instead of listing every possible cause. Each base alert rule adds the
   # matching query as annotation-only `Info`; it is not part of the threshold
--- a/alerts/rules/rules-fpmms.tf
+++ b/alerts/rules/rules-fpmms.tf
@@ -705,6 +705,8 @@ resource "grafana_rule_group" "fpmms_deviation" {
 #              the pool still serves both directions.
 #   page     — min value share < 10%. The pool is effectively one-sided;
 #              swappers lose one direction outright once it empties.
+#   warning  — the pool holds reserves but publishes no value share at all, so
+#              neither tier above can see it. Coverage visibility, not a band.
 #
 # Neither tier fires on the 2026-08 CHFm/USDm breach that motivated ADR 0067:
 # that pool bottomed at ~22% min side, above both bands. `Rebalancer Stale`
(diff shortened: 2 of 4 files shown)


chapati23:
Won't fix: this is an infrastructure failure in the `auto-review` job, not a finding against the diff. Both the original run and the re-run I triggered return `{"subtype":"success","is_error":true,"duration_ms":441,"num_turns":1,"total_cost_usd":0}` — the model call fails immediately at zero cost, so the action never reached the diff. Nothing in this PR can change that outcome; `auto-review` is also not one of the branch-protection required checks.

Review coverage for this change came from the other reviewers, all clean on `544806fb0e339c7e2dd91716c7a47066163087e3`:

- Local codex autoreview: `autoreview clean: no accepted/actionable findings reported` — "overall: patch is correct (0.89)".
- CodeRabbit: "No actionable comments were generated in the recent review", merge risk Minimal.
- Cursor Bugbot: pass.
- `Terraform Plan (alerts/rules)`: PR-targeted secretless plan clean.

Re-run the job once the action's credentials recover if a fresh pass is wanted.

PR #2119, opened 2026-08-28
test(docs): lock historical navigation validation semantics

### LINKED ISSUE #2104: fix(docs): navigation-eval enforces the forward-looking headroom reserve against historical inventories, so a reserve fix cannot go green on its own PR
## Context

`docs-navigation-eval.mjs --validate` rebuilds the documentation inventory at
the result's `repository_base_commit` and then re-runs the full fixture
contract against that **historical** corpus, including the forward-looking
reserve `min_total_unique_source_headroom_bytes`.

`scripts/docs/docs-navigation-eval.mjs`:

```js
const inventory = inventoryCommit
  ? buildHistoricalNavigationInventory({ suite, repoRoot, commit: inventoryCommit })
  : buildDocumentationInventory({ repoRoot, files: trackedDocumentationFiles(repoRoot) });
...
const fixtureErrors = validateFixtureSuite(suite, inventory, {
  // The immutable pre-garden contract predates the live reserve target.
  requireHeadroomReserve: fixturesPath !== baselineFixturesPath,
});
```

`scripts/docs/docs-navigation-eval.test.mjs` pins its synthetic result to
`git merge-base HEAD origin/main`, so the test "CLI checks fixtures and
validates a structured result" measures the reserve against the **merge-base
commit**, not the working tree.

## What this causes

The reserve becomes unrecoverable from a branch. While `main` is underwater,
a PR that genuinely restores the reserve still fails
`pnpm docs:navigation-eval:test`, because the merge-base corpus it validates
against is the unfixed `main`.

Measured on the #2102 fix branch (two routed sources trimmed by 1,782 bytes,
live headroom restored to 34,218):

| merge-base `origin/main` | `docs:navigation-eval:test` |
| ------------------------ | --------------------------- |
| `6b39e304` (unfixed)     | 1 of 34 fails — "cheapest accepted route union leaves 32436 bytes of headroom" |
| the fix commit itself    | 34 of 34 pass |

The second row was produced by pointing the local `refs/remotes/origin/main`
at the fix commit and rerunning, then restoring the ref. So the fix is
correct and goes green the moment it is on `main` — but it cannot prove that
in its own CI run.

A second consequence is independent of that bootstrap: a committed monthly
result can become invalid later purely because `main`'s corpus grew, with no
change to the recorded run it is supposed to score.

## Why it may be wrong to enforce the reserve here

The reserve is a forward-looking capacity target — it reserves room for the
*next* documents. A past commit's corpus cannot be changed, and `--validate`
scores a recorded evaluation run, not the repository's future headroom. The
same function already disables the reserve for the frozen baseline contract
for a closely related reason ("The immutable pre-garden contract predates the
live reserve target").

## Proposed option (needs a human decision)

Do not require the headroom reserve when the inventory is historical
(`inventoryCommit` set), while keeping every other fixture rule and the hard
`max_total_unique_source_bytes` cap enforced there.

This would not weaken the live control: the reserve is enforced by
`--check-fixtures` against the working tree, which runs in CI twice
(`ci.yml` "Documentation navigation fixture contract" in both the
`Documentation corpus checks` and `Lint + test root scripts` jobs) and in the
`agent-quality-gate-pre-push` hook whenever a diff touches `docs/`.

The decision is whether the reserve is meant to be a property of the live
tree only, or also a property every historical evaluation must satisfy.
That is a control-semantics call, so this is filed for grooming rather than
changed inside the #2102 hotfix.

## Files

- `scripts/docs/docs-navigation-eval.mjs` (inventory selection and
  `requireHeadroomReserve`)
- `scripts/docs/docs-navigation-eval-helpers.mjs` (`validateFixtureSuite`)
- `scripts/docs/docs-navigation-eval.test.mjs` (the merge-base pin)

## Done means

- A branch that restores the reserve passes `pnpm docs:navigation-eval:test`
  in its own CI run, without weakening the live `--check-fixtures` reserve.
- A committed evaluation result stays valid when the live corpus later grows.
- The chosen semantics are stated where the eval is documented
  (`docs/evals/documentation-navigation.md`).

chapati23:
## The Problem

- Historical navigation-result validation enforced the forward-looking live headroom reserve against the result's pinned inventory. A branch that restored live headroom could still fail against its merge-base inventory.
- A recorded result could also become invalid after later documentation growth. Existing tests did not prove the reserve and hard-cap boundary behavior through the evaluation loader and CLI.

## The Solution

This PR documents and pins the split validation contract. Live fixture checks enforce the headroom reserve. Historical result checks skip that reserve but still enforce the hard `max_total_unique_source_bytes` cap. The regression test covers exact equality and the adjacent failing value for both limits through `loadEvaluationContext`, then validates the recorded result through the CLI path.

The production predicate already exists on `main` from #2075. This PR adds the missing contract evidence and evaluator documentation. It does not change route scoring, source inventories, or budget values.

## Details

- Build a committed historical inventory in an isolated temporary Git repository.
- Prove that live validation accepts reserve equality and rejects a reserve that is one byte too large.
- Prove that historical validation skips that reserve, accepts hard-cap equality, and rejects a hard cap that is one byte too small.
- State the live and historical rules in the navigation evaluation document.

## Validation

- `pnpm docs:navigation-eval:test` passed all 34 tests.
- `pnpm docs:navigation-eval -- --check-fixtures --json` passed with a 227,210-byte floor, 34,790 bytes of headroom, and a 2,022-byte reserve surplus.
- `pnpm agent:quality-gate --run` passed all 8 mapped commands, including Trunk, documentation index and context checks, script lint, live and historical navigation checks, and Terraform tests.
- These checks prove the deterministic fixture, loader, and CLI contract. They do not prove model answer quality or future documentation size.

## Deferrals

- None

## Checklist

- [x] Architecture decision? Not applicable. This PR documents and tests the existing validation boundary.

Closes #2104

---

> [!NOTE]
> **Low Risk**
> Docs and test-only changes that document and pin existing loader validation; no production logic or budget values are modified.
> 
> **Overview**
> Documents and regression-tests the **split byte-budget rules** for documentation navigation evaluation: live `--check-fixtures` still enforces both `max_total_unique_source_bytes` and the forward-looking `min_total_unique_source_headroom_bytes` reserve against the **current** inventory, while historical `--validate` **skips only the headroom reserve** but still applies the hard suite cap against the inventory at the result’s pinned `repository_base_commit`.
> 
> The new cases in the isolated-git historical test drive `loadEvaluationContext` through **exact boundary values**—reserve at the limit vs one byte too high on live load, then the same reserve violation allowed when `inventoryCommit` is set, plus hard-cap equality vs one byte under the pinned floor—before restoring the fixture and continuing the existing CLI historical validation path. **No evaluator logic or fixture budgets change** in this diff; it adds contract text and evidence for behavior already on `main`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdde5aca658fb17f6f840af3f70282afb33cafae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Documentation**
  * Clarified validation behavior for historical and current documentation-navigation evaluations.
  * Documented how headroom and total-source-byte limits are applied.

* **Tests**
  * Added boundary checks for validation limits, including minimum passing values and over-limit failures.
  * Added coverage for loading historical inventory data.

### DIFF
--- a/docs/evals/documentation-navigation.md
+++ b/docs/evals/documentation-navigation.md
@@ -60,6 +60,11 @@ failing committed baseline before treating it as historical comparison
 evidence. They validate it against the frozen baseline fixtures, not the
 current contract.
 
+Historical `--validate` runs skip only the forward-looking total-source
+headroom reserve. They still enforce `max_total_unique_source_bytes` against
+the inventory at `repository_base_commit`. The live `--check-fixtures` run
+enforces both limits against the current documentation inventory.
+
 `sources_requiring_verification` entries are historical qualification traps,
 not live routes. A trap may remain in the immutable fixture after its document
 is retired only when the evaluator explicitly lists that path as a tombstone;
--- a/scripts/docs/docs-navigation-eval.test.mjs
+++ b/scripts/docs/docs-navigation-eval.test.mjs
@@ -721,6 +721,91 @@ test("historical scoring requires a default-branch ancestor and survives deletio
       cwd: temp,
       encoding: "utf8",
     }).trim();
+    const pinnedContext = loadEvaluationContext(
+      {
+        repoRoot: temp,
+        fixturesPath: fixturePath,
+      },
+      { inventoryCommit: commit },
+    );
+    const pinnedFloor = navigationContextFloor(
+      pinnedContext.suite,
+      pinnedContext.inventory,
+    );
+    const reserveAtLimit = structuredClone(context.suite);
+    reserveAtLimit.targets.min_total_unique_source_headroom_bytes =
+      reserveAtLimit.targets.max_total_unique_source_bytes -
+      pinnedFloor.total_unique_route_bytes;
+    writeFileSync(
+      path.join(temp, fixturePath),
+      `${JSON.stringify(reserveAtLimit)}\n`,
+    );
+    assert.doesNotThrow(() =>
+      loadEvaluationContext({
+        repoRoot: temp,
+        fixturesPath: fixturePath,
+      }),
+    );
+    const reserveTooHigh = structuredClone(reserveAtLimit);
+    reserveTooHigh.targets.min_total_unique_source_headroom_bytes += 1;
+    writeFileSync(
+      path.join(temp, fixturePath),
+      `${JSON.stringify(reserveTooHigh)}\n`,
+    );
+    assert.throws(
+      () =>
+        loadEvaluationContext({
+          repoRoot: temp,
+          fixturesPath: fixturePath,
+        }),
+      /cheapest accepted route union leaves .* bytes of headroom/,
+    );
+    assert.doesNotThrow(() =>
+      loadEvaluationContext(
+        {
+          repoRoot: temp,
+          fixturesPath: fixturePath,
+        },
+        { inventoryCommit: commit },
+      ),
+    );
+    const hardCapAtLimit = structuredClone(reserveTooHigh);
+    hardCapAtLimit.targets.max_total_unique_source_bytes =
+      pinnedFloor.total_unique_route_bytes;
+    writeFileSync(
+      path.join(temp, fixturePath),
+      `${JSON.stringify(hardCapAtLimit)}\n`,
+    );
+    assert.doesNotThrow(() =>
+      loadEvaluationContext(
+        {
+          repoRoot: temp,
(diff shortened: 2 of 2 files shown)


chapati23:
@coderabbitai review

Issue #2253, opened 2026-09-03
Code Quality: audit steps before Run Trunk hide lint results


chapati23:
## Problem

`Code Quality` (`.github/workflows/trunk.yml`, job `trunk`) runs six steps before
`Run Trunk`, and none of them uses `if: always()`:

1. `Check GitHub Actions pins`
2. `PR validation boundary suite`
3. `pnpm audit high gate self-tests`
4. `pnpm audit governance-watchdog deploy lockfile (high+ only)`
5. `pnpm audit on-call announcer deploy lockfile (high+ only)`
6. `pnpm audit handler deploy lockfile (high+ only)`

A failure in any of them aborts the job before `./tools/trunk check --ci --all`
executes. The check reports red, but the lint result is unknown, so a lint
regression can sit on `main` undetected for as long as an earlier step stays red.

That is not hypothetical. The four `fast-uri` advisories (#2248) held the
`governance-watchdog` audit step red from the time PR #2228 merged. PR #2249
cleared the advisories; `Run Trunk` then ran for the first time since #2228 and
reported six shellcheck findings that had been on `main` the whole time
(#2252, #2251). Two independent defects were stacked behind one red check, and
the second only became visible when the first was fixed.

## Solution ideas

Options, from smallest to largest:

- Give `Run Trunk` `if: always()` (or `if: '!cancelled()'`) so lint runs whatever
  the earlier steps did. The job still fails; the difference is that both results
  are reported in the same run.
- Move the supply-chain audit steps after `Run Trunk`. This keeps the fail-fast
  behaviour for pins and the validation-boundary suite, which are cheap.
- Split the audits into their own job in the same workflow, so `Code Quality`
  reports lint and the audits as independent results.

Decide whether the required-check contract expects one context or several before
picking. `docs/pr-checklists/ci-workflow-gates.md` lists `Code Quality` as one
ruleset-required context, so splitting jobs needs a ruleset change in the same
PR.

## Done means

An earlier step's failure in the `Code Quality` job cannot hide the result of
`./tools/trunk check --ci --all`, and the required-status contract is still
satisfied on every PR.

chapati23:
Agent claim: codex claimed #2253 for implementation.

Claim ID: claim-f53c2e6e-972f-455e-995f-c67947aaaa5b
Branch: fix/2253-trunk-independent-lint
Claimed at: 2026-09-05T17:33:11.099Z

chapati23:
[...]

PR #2249, opened 2026-09-02
fix(deps): pin fast-uri to 3.1.6 for four new high advisories

chapati23:
## The Problem

- The required `Code Quality` check was red on `main` and on every open PR, blocking merges repo-wide for reasons unrelated to any PR's diff. **Two independent defects were stacked behind it**, and the second was invisible: the job runs its `pnpm audit` steps before `Run Trunk`, so it died at the first audit step and never reached Trunk.
- Four high advisories published 2026-09-02 land on `fast-uri` 3.1.5: [GHSA-5jgf-p345-68v8](https://github.com/advisories/GHSA-5jgf-p345-68v8) (host confusion via skipped IDN canonicalization), [GHSA-f65p-4m7j-42xc](https://github.com/advisories/GHSA-f65p-4m7j-42xc) (SSRF via malformed IPv6 normalization), [GHSA-fph4-wmhf-6fwf](https://github.com/advisories/GHSA-fph4-wmhf-6fwf) (SSRF via repeated hostname percent-decoding), and [GHSA-jqff-g426-hqxp](https://github.com/advisories/GHSA-jqff-g426-hqxp) (host confusion via percent-encoded scheme normalization). `ajv` pulls `fast-uri` into all four dependency roots.
- Every root already carried `"fast-uri@<3.1.5": 3.1.5`, added to escape GHSA-7p8r-x3mc-p8w7. That clause is satisfied by 3.1.5 itself, so the override actively held the graph on the version the four new advisories hit. This was a stale override, not a missing one.

## The Solution

Raise the existing clause to `"fast-uri@<3.1.6": 3.1.6` in all four dependency roots and refresh all four lockfiles, so every root resolves `fast-uri@3.1.6` — the first patched release on the 3.x line for all four advisories. The repo stops shipping a URI parser with two SSRF and two host-confusion holes into the Cloud Functions that parse webhook payloads.

The clause is **replaced**, not supplemented: a second `<3.1.6` override alongside the old `<3.1.5` one would leave the two fighting over the same package. The floor stays on the 3.x line because `ajv` 8, the only consumer here, depends on `fast-uri` `^3`; npm `latest` is 4.1.4 and moving there would be a major bump of a transitive dependency with no advisory benefit.

[...]

Issue #2251, opened 2026-09-02
Code Quality still red after the fast-uri fix: six shellcheck findings in agent-quality-gate.test.sh


chapati23:
## Goal

Clear the six shellcheck findings in `scripts/agent-quality-gate.test.sh` so the
ruleset-required `Code Quality` check can go green once the `fast-uri` fix
lands. Fixing `fast-uri` alone is not enough: a second, independent failure sits
behind it.

## Context and links

`Code Quality` runs its three `pnpm audit` steps before `Run Trunk`. While
`main` resolves the vulnerable `fast-uri@3.1.5`, the job dies at the first audit
step and never reaches Trunk, so this failure is invisible on `main`.

PR 2249 fixes `fast-uri`. Its `Code Quality` run therefore got past the audit
steps and reached `Run Trunk`, which failed
(run 33684909673, head `13c6a573`):

```
scripts/agent-quality-gate.test.sh:6600:5
 6600:5  low  This function is never invoked. Check usage (or ignored if invoked indirectly).  shellcheck/SC2329
 6604:5  low  This function is never invoked. Check usage (or ignored if invoked indirectly).  shellcheck/SC2329
 6669:7  low  Modification of gate_run_id is local (to subshell caused by (..) group).         shellcheck/SC2030
 6688:7  low  This function is never invoked. Check usage (or ignored if invoked indirectly).  shellcheck/SC2329
 6694:7  low  This function is never invoked. Check usage (or ignored if invoked indirectly).  shellcheck/SC2329
 6695:7  low  This function is never invoked. Check usage (or ignored if invoked indirectly).  shellcheck/SC2329
Checked 2486 files
✖ 6 lint issues
```

The last green `Code Quality` on `main` was run 33650631209 at 2026-09-02
15:45:36 UTC on `5a37407a`. Two commits touched this file after that point, both
merged while the audit step was already masking Trunk:

```
$ git log --oneline 5a37407a..origin/main -- scripts/agent-quality-gate.test.sh
5b3c21a2 fix(gate): arm the drain refresh barrier on the named drain only (#2228)
a33f7ca3 Remove obsolete review metrics collector (#2243)
```

`git blame` on `origin/main` attributes all six flagged lines to `5b3c21a2`.

This is the "broken controls come in chains" pattern: one red control on a clean
`main` hid a second defect, so sweep the whole `Code Quality` job rather than
stopping at the first green step.

## Acceptance criteria

[...]

PR #2254, opened 2026-09-03
fix(gate): mirror the SC2030 waiver onto the Linux fixture and correct two teardown comments

chapati23:
## The Problem

- `main` now carries the shellcheck fix for the gate test suite, folded into
  a0f2cde9 (#2249). Two defects in what that fix added survived the fold.
- **The SC2030 waiver is positional.** shellcheck reports one SC2030 per
  variable. It currently lands on the Darwin `gate_run_id` write, which carries
  the `disable` — so the byte-identical write in the Linux fixture is quiet only
  because the Darwin one absorbs the finding, and has no directive of its own.
  Delete or reorder that Darwin block and the finding moves to the unprotected
  line, turning the ruleset-required `Code Quality` check red again.
- **One comment names a mechanism that does not happen.** It says the teardown
  "returns before its descendant walk" on the Linux path. It does not return
  there; a maintainer trusting that sentence would conclude nothing after the
  guard runs.

## The Solution

The `# shellcheck disable=SC2030` directive now sits on **both** `gate_run_id`
fixture writes, so the suppression no longer depends on which of the two
shellcheck happens to attribute the finding to. Either fixture can be edited,
reordered, or removed without re-reddening `Code Quality`. The two inaccurate
comments are corrected to describe what the code actually does.

No behaviour change: this is three comment blocks and one added directive. The
test's assertions, its fixtures, and every runtime gate file are untouched.

The material limit is unchanged from the original fix — the directive suppresses
SC2030 at those two lines only, and shellcheck still cannot see through the
`eval` that hides the lifted teardown's call graph. A genuinely dead stub added
next to these would still not be flagged.

## Details

This branch is `main` plus this residual: it merges a0f2cde9, and
`git diff origin/main...HEAD` is one file at 15 insertions / 6 deletions.

**`roots`, not an early return.** `collect_process_tree` goes uncalled in the
Linux fixture because the teardown's descendant walk iterates `roots`, which is
built only from `active_timeout_records` — empty in that fixture — so the loop
runs zero times. The walk is *reached*: the guard at
`scripts/agent-quality-gate.sh:1508-1513` returns only when every registry is
empty, and the fixture sets `active_worker_pgids=(4000001 4000002)`. The comment
now says that.

**Darwin restore guidance.** The Darwin comment told a future maintainer to
restore a `collect_process_tree` stub alongside the terminate stub. That path
returns at `scripts/agent-quality-gate.sh:1679` before the walk whether or not
records are registered, so only the terminate stub could ever be needed there;
restoring the other would re-raise SC2329. The comment now scopes the advice to
the one stub that could matter.

## Validation

- `git diff --no-ext-diff origin/main...HEAD --stat` — one file,
  `scripts/agent-quality-gate.test.sh`, 15 insertions / 6 deletions. Nothing
  else differs from `main`.
- `shellcheck -x` — clean, exit 0. `./tools/trunk check --ci
  scripts/agent-quality-gate.test.sh` — `Checked 1 file`, `✔ No issues`.
- **Negative control, both directions, run on this merged tree.** Delete the
  Darwin `gate_run_id` write and:
  - with this PR's directive present → shellcheck clean, exit 0;
  - with the directive removed → `SC2030` reported at the Linux write, exit 1.

  So the added directive does real work: it is the thing standing between a
  future edit of the Darwin fixture and a red required check. This proves the
  suppression is no longer positional; it does not prove anything about other
  shellcheck rules in the file.
- Not re-run here: the full `pnpm agent:quality-gate:test` suite. It passed on
  this branch's parent content at `4710aa46` (run 33720164275) before the fold,
  and this residual changes comments plus one directive, so CI on this head is
  what establishes it.

## Deferrals

- #2253 — the `Code Quality` job runs six steps before `Run Trunk` with no
  `if: always()`, so an earlier failure hides the lint result entirely. This is
  what made the original findings invisible on `main`, and it also made the two
  fixes mutually unmergeable until they were folded together. Reordering or
  splitting those steps changes a required-status contract, so it is not done
  here.

Refs #2253

https://claude.ai/code/session_01McsUkrwYBL6oC7CdZCCsc4


chapati23:
Status check against current `main` (`a0f2cde9`), since the situation moved: **this PR is not superseded, but it is no longer blocking anything.**

#2249 merged at 06:36 carrying both halves — the `fast-uri` pin *and* a version of the shellcheck fix (`scripts/agent-quality-gate.test.sh`, +22/-3). `Code Quality` is green on `main` at `a0f2cde9`, so the deadlock is resolved and #2251 is closed.

What this branch (`312b4302`) still adds over `main`, from `git diff origin/main origin/fix/2252-gate-test-shellcheck`:

1. **A real hardening.** The `# shellcheck disable=SC2030` directive on the Linux fixture's `gate_run_id` write. `main` carries it only on the Darwin block. shellcheck reports one SC2030 per variable, so the directive currently lands there by position — removing or reordering the Darwin fixture would silently re-red `Code Quality`. This makes it position-independent.
2. **Two comment corrections.** `main`'s comments say the teardown "returns before its descendant walk". This branch corrects that to: the walk *is* reached, and never iterates because `roots` is built only from `active_timeout_records`, which is empty — the preceding guard returns only when every registry is empty, and `active_worker_pgids` is not. Same correction for the Darwin block. The `main` wording would mislead anyone deciding whether it is safe to keep the stub deleted.

Both are worth keeping; neither is urgent. I have not touched this PR and am not closing it — the merge decision is the operator's.

PR #2257, opened 2026-09-03
fix(tooling): read mutex refs via REST and snapshot the board once per sync

### LINKED ISSUE #2226: tooling: issue:review always fails on a claimed issue because GraphQL cannot read the custom mutex ref
## Goal

Make `pnpm issue:review` work on an issue whose retained mutex ref already
exists. Today it always fails, because the helper reads the mutex through
GitHub GraphQL and GraphQL does not resolve refs in the custom lock namespace.

## Context and links

`docs/notes/github-tooling-surfaces.md` states the design: "The mutex uses
GraphQL `updateRefs` with an exact `beforeOid` on a retained custom ref."
`readLockRef` in `scripts/pr/issue-board-lock.mjs:906` reads it with
`repository.ref(qualifiedName:)`. That query returns `null` for the lock
namespace, so `initializeLockRef` cannot reconcile after its create-from-absent
compare-and-swap loses to the ref that is already there. It retries three times
and throws `scripts/pr/issue-board-lock.mjs:1132`.

Observed 2026-09-02 while moving issue 2221 to review for PR 2225:

```text
$ pnpm issue:review --pr 2225 --issue 2221
Issue #2221 mutex ref refs/mento-issue-board-locks/v1/753bd01051bdacf8d22ebd0f44dc2275f3762563632938d0a2d7367d5da1d3b9 was absent after 3 compare-and-swap attempts
```

The ref is not absent. `git ls-remote` resolves it:

```text
93a1a4612331f5ea57d136e2a80fbc07d168ff35	refs/mento-issue-board-locks/v1/753bd01051bdacf8d22ebd0f44dc2275f3762563632938d0a2d7367d5da1d3b9
```

REST resolves it too:

```text
$ gh api repos/mento-protocol/monitoring-monorepo/git/ref/mento-issue-board-locks/v1/753bd010…
{"ref":"refs/mento-issue-board-locks/v1/753bd010…","object":{"sha":"93a1a4612331f5ea57d136e2a80fbc07d168ff35","type":"commit"}}
```

GraphQL does not:

```text
$ gh api graphql -f query='query($owner:String!,$name:String!,$ref:String!){repository(owner:$owner,name:$name){id ref(qualifiedName:$ref){id target{__typename oid}}}}' …
{"data":{"repository":{"id":"R_kgDORdux6w","ref":null}}}
```

This is not specific to issue 2221's ref. A second, older lock ref from the same
namespace also reads `null` through GraphQL while `git ls-remote` lists it:
`refs/mento-issue-board-locks/v1/0350094961cb1e9c930109998cc150970f07696f34bfd7dbd3cfa1cbdc6c556d`.

`pnpm issue:claim` still works, because it creates the ref when the read reports
absent and there is nothing yet to reconcile. Every transition that must read an
already-retained ref is affected, so `issue:release` and any second transition
on the same issue are likely to fail the same way. That is worth confirming as
part of the fix.

## Acceptance criteria

- [ ] `pnpm issue:review --pr <pr> --issue <n>` succeeds on an issue whose lock
      ref already exists.
- [ ] The mutex read no longer depends on GraphQL resolving a custom ref
      namespace, or the helper falls back to the Git data REST endpoint that does
      resolve it.
- [ ] `issue:release` and a repeated transition on one issue are checked for the
      same failure and fixed together.
- [ ] A regression test covers the reconcile path with the ref already present.
- [ ] The compare-and-swap safety property is unchanged: a competing writer must
      still lose. Do not replace the mutex with an unconditional write.

## Expected files or package area

- `scripts/pr/issue-board-lock.mjs` (`readLockRef`, `initializeLockRef`,
  `reconcileLockRefRead`)
- `scripts/pr/agent-issue-board.mjs`
- `docs/notes/github-tooling-surfaces.md` if the transport changes

## Verification commands

```bash
node scripts/pr/agent-issue-board.test.mjs
pnpm issue:review --pr <pr> --issue <n>
```

## Risks and do-not-touch

- Do not weaken the mutex to get a transition through. The compare-and-swap on
  an exact `beforeOid` is the control that stops two agents owning one issue.
- Do not delete or rewrite lock refs by hand as the fix.
- The error text says "was absent" when the ref is present. Fix the message too;
  it sent this investigation the wrong way first.

## Done means

Board transitions work on already-claimed issues, the mutex still refuses a
competing writer, and the failure message names the real condition.

Refs #2221, #2225.

chapati23:
## The Problem

- Every issue-board mutex operation on an issue whose retained lock ref already existed failed with "mutex ref … was absent after 3 compare-and-swap attempts". GitHub GraphQL `repository.ref(qualifiedName:)` returns null for refs outside `refs/heads` and `refs/tags`, so the helper never saw the ref it had created earlier. Since 2026-08-31 no lock ref has recorded a second operation: `issue:review` and `issue:release` fail after any claim, and `issue:board sync` failed on every closed issue that still carried a queue label.
- `issue:board sync` spent two `gh` calls on each of ~80 listed issues to decide that a stable issue needed nothing, then burned another dozen calls per closed issue in the doomed lock initialization. A full run took 5m45s and ended with 14 failures.

## The Solution

- The mutex reads its ref through the REST matching-references endpoint and the ref's commit through GraphQL `repository.object`. Both resolve the custom namespace. The compare-and-swap path, the exact `beforeOid` precondition, and the retained ref chain are unchanged. The "absent" error now names the last compare-and-swap failure.
- The board enumeration attaches exact Project item and Project node IDs with one batched `nodes(ids:)` read per 100 issues and marks the snapshot complete. The sync preflight trusts that complete snapshot, so a stable open issue costs no per-issue read. Every mutation still re-reads under the mutex. An issue that changes after enumeration is handled by the next run, the same deferral the grooming check already documents.
- Live result on the 78-issue board: 2m18s for the run that cleaned the 13 stale closed issues, then 7s in steady state, with only the two intended grooming findings.
- Non-goal: no concurrency across issues and no change to the mutex protocol.

## Details

- `readLockRef` (`scripts/pr/issue-board-lock.mjs`) calls `GET repos/{repo}/git/matching-refs/mento-issue-board-locks/v1/<digest>`, keeps only the exact `ref` match, and rejects a non-commit target before reading the commit. It is exported with an injectable `{ json, graphql }` transport for tests. The unused ref node `id` is dropped from the record.
- `ghGraphql` accepts array variables and emits them as repeated `-f key[]=value` fields.
- `listIssuesByLabels` (`scripts/pr/issue-board-transport.mjs`) keeps the `gh issue list` call and adds `attachExactProjectItems`. An incomplete or missing page leaves that issue's title-only snapshot without the completeness marker, so `findIssueProjectItem` and the preflight fall back to a per-issue read as before.
- `preflightStableSyncIssue` (`scripts/pr/issue-board-sync-lock.mjs`) uses the listed issue when `projectItemsPageInfo.hasNextPage === false`, otherwise it reads the issue as before.
- `githubProjectScopeHint` treats `git/matching-refs` as a lock API path.
- ADR 0082 and `docs/notes/github-tooling-surfaces.md` now state the REST read and why GraphQL cannot observe the namespace.

## Validation

- `node scripts/pr/agent-issue-board.test.mjs`: 260 passed (257 before, plus a REST ref-read test, a batched exact-item test, and a zero-read preflight test). The suite proves the new read path against fake transports and the reconcile path against the fake lock server; it does not exercise the live GitHub API.
- Live, before the fix: `gh api graphql` with `ref(qualifiedName:)` returned `{"ref":null}` for the existing 2245 lock ref; `git ls-remote` and `gh api .../git/ref/...` resolved it. All 72 lock refs held exactly one operation (18 claim, 2 review, 52 sync initializations), none a second one.
- Live, with the fix: `pnpm issue:claim --issue 2226` acquired and released the mutex on 2226's pre-existing ref (rejected as not claimable because another session had claimed it seconds earlier, which is the expected outcome). That other session's `issue:release` then failed with the old code's "was absent" error. `pnpm issue:review --pr 2257 --issue 2226` with this branch succeeded on the same retained ref, the first acceptance criterion of the issue; the ref now ends in an UNLOCK whose payload records `pr=2257` on top of the earlier claim LOCK and UNLOCK. `pnpm issue:board sync --dry-run` 30s; `pnpm issue:board sync` 2m18s with 76 successes and the 2 grooming findings, clearing the stale labels on the 13 closed issues; a second run 7s. These runs prove the read and the speedup on this board; they do not prove behavior under concurrent syncs from two machines.
- `pnpm exec eslint` on the four changed scripts: clean. `trunk fmt` and the commit-time `trunk check` on the six files: clean.
- The local pre-push gate did not run. Two `--run` attempts exited 2 before executing any mapped command: first a coordinator serving from the deleted `coderabbit-migration-closeout` worktree, then a stale `cmd…-80232` holder record with missing process-lineage evidence (the wedge in issues 2201 and 2216). The operator instructed in session: "fuck the gate, skip it locally", so the push used `--no-verify`. CI runs the same mapped commands on this PR; treat the PR checks as the gate evidence.

## Deferrals

- None

## Checklist

- [x] Architecture decision? ADR 0082 updated in place to record the transport fact; no new decision.

Closes #2226.

https://claude.ai/code/session_01SEXK2xfxvctqZbWAVTqfY2


chapati23:
Both review points are informational and neither changes behavior. Dispositions below.

**Won't fix — `githubProjectScopeHint` matching-refs test coverage.** The gap is real but cosmetic. The new alternative feeds the same lock-API branch that `mutex GraphQL failures name the Contents write requirement` already asserts through `updateRefs`. Rather than assume the alternation works, I exercised it directly against the module:

```
matching-refs   -> Contents-write guidance: true
refs            -> Contents-write guidance: true
commits         -> Contents-write guidance: true
unrelated       -> Contents-write guidance: false
```

The unrelated case is the negative control: the branch still discriminates, so the behavior the suggested test would assert is already shown correct. Committing the assertion needs a push, and a push invalidates the current-head Codex PR-description approval, which is a required gate on this repo. I am not trading a satisfied required sign-off for an assertion on verified behavior. Happy to add it if a maintainer wants the assertion in the suite.

**Won't fix — `readLockRef` issues the GraphQL read before rejecting a non-commit REST object type.** Agreed with the assessment. `issueMutationLockRef` only ever creates commit-target refs under `refs/mento-issue-board-locks/v1/<digest>`, so reaching that path takes a hand-made annotated tag at the lock ref, and the validation immediately after still raises the same conflict with the same scope and ref. Short-circuiting would add a branch for a state this code cannot produce.

Local validation at `f3f2ce2`: `node scripts/pr/agent-issue-board.test.mjs` — 260 passed.

chapati23:
Correction to my previous reply: the matching-refs assertion is in after all.

The reason I gave for declining — that a push would cost the current-head Codex PR-description approval — stopped applying once I read the ruleset. `main` runs `strict_required_status_checks_policy: true`, so this branch had to be brought up to date before it could merge at all. A push was required either way, which made the test free to include.

`eb5b3c62` adds the case: a REST matching-refs argument must produce the Contents-write guidance, paired with a negative control on an unrelated API path so the branch is shown to still discriminate. Dropping the `matching-refs` alternation fails that test and nothing else, so it reaches live code rather than passing vacuously.

`7806ff2c1` merges `origin/main` and clears the behind-base state.

The second point stands as won't-fix, for the reason given above it.

Validation at `7806ff2c1`: `node scripts/pr/agent-issue-board.test.mjs` — 261 passed.

PR #2485, opened 2026-09-17
fix(tooling): probe gh for --attach in both hosted setups

chapati23:
## tl;dr

Agents changing the dashboard from a hosted container could not attach the
before/after screenshots a review needs, and found out only at the end, with the
work done. Both hosted setups now check that ability at startup, fix it where
they can, and name what to unblock otherwise. Nothing changes locally.

## The Problem

- The dashboard verification runbook uploads visual evidence with
  `gh pr edit --attach`, added in gh 2.99.0.
- Both hosted bootstraps accepted any `gh`; Codex Cloud kept the distro package.
- A container on gh 2.98.x reached publication before the gap surfaced.

## The Solution

- Both probe the flag in captured `gh pr edit --help` text, not a parsed version:
  distro builds backport flags unevenly.
- The web setup adds it to the condition that triggers its existing
  release-tarball install, and warns whenever this container cannot upload at
  all, MCP fallback included.
- Codex Cloud upgrades through an existing GitHub CLI apt source, or adds the
  official one, then reports a precise blocker. Never fatal: only dashboard
  evidence needs the flag.

## Details

- Done means read as: both setups detect a `gh` without `--attach` and either
  upgrade or report a precise blocker, with tests.
- Simplest version: a probe plus a message. The rest guards apt state, since the
  installer can now run where apt already works: it stops at the first failed
  step, stages the keyring through a temporary file, removes a source it added
  when no usable `gh` results, and never rewrites an existing one.
- `codex-cloud-github-cli.sh` takes the moved provisioning, dropping the entry
  point 608 → 564 lines rather than growing past a cap it already exceeded.
- Won't fix here: running the gh capability gate's live REST and GraphQL probes
  at setup. That contract is shared with `.claude/babysit-pr.sh`.

## Validation

Head `82011be3`. Gates: `agent-setup-contract.test.sh` ✓,
`codex-cloud-setup.test.sh` 23 cases ✓, `ci:contract:test` 189 ✓,
`shellcheck 0.11.0` + `bash -n` 5 files ✓, `lint:scripts` ✓,
`docs:index --check` ✓, `docs:navigation-eval:test` 37 ✓,
`agent:context-check` ✓, `agent:context-budget --strict` + `:test` 17 ✓,
action pins ✓, autofix trust ✓, `trunk fmt` ✓.
Review: `agent:closeout-review` ×7 → 11 findings, 10 fixed, 1 declined above.
Not run: Trunk's shellcheck — its cached binary is x86_64 on this arm64 host;
the pinned 0.11.0 on PATH ran instead, clean.
Claude Security scan: skipped (apt and keyring handling); not declared here.
Scope baseline: merge base `a7d31b95`, 4 files, 90 non-test added lines; final
10 files, 219, of which 45 are the move — over the 180 threshold, reported.
Not proven: no hosted container ran this; every case uses a mocked installer.

## Deferrals

- None

## Summary by CodeRabbit

- **New Features**
  - Hosted environments now check whether the GitHub CLI supports attaching visual evidence to pull requests.
  - GitHub CLI installation and upgrades can use official package sources when required capabilities are unavailable.
  - Setup continues after unsuccessful upgrades while clearly warning when visual-evidence support remains unavailable.

- **Bug Fixes**
  - Improved GitHub CLI capability detection and fallback handling for unsupported or unauthenticated installations.
  - Added safeguards to avoid leaving behind temporary package sources after failed upgrades.

- **Tests**
  - Expanded coverage for GitHub CLI installation, upgrades, capability detection, and failure scenarios.


chapati23:
@coderabbitai review

